Issue #66, opened 2025-03-19
Event-based feature extraction


PhilippMatthes:
Currently the feature extractor runs in a 1-minute routine without knowledge of the syncer. However, it would make sense to re-calculate features directly after the underlying data has changed. 

1. The syncer needs to notify the feature extractor somehow (e.g. via mqtt) that a new sync has happened.
2. Use the dependency graph already provided in the config to see which features need re-calculation.

PR #67, opened 2025-03-20
Event-based feature extraction (MQTT)

PhilippMatthes:
Adds event-based feature extraction to Cortex: features are now extracted as soon as the underlying data was updated. 

Closes: #66 

- If mqtt is enabled, the syncer will now publish triggers there.
- Triggers are received by the feature extractor to re-trigger feature extraction on-demand.
- Don't re-calculate all features. Calculate distinct feature extractor subgraphs to determine which features need re-calculation.
- Refactor the config a bit to make it more consistent with the newly introduced triggers.
- Wrap the `Publish` function in the mqtt client to not write `//nolint` everywhere.
- Add architecture decision record
- Make mqtt broker deployment mandatory
- Disable periodic feature extraction

Co-authored-by: mblos <marcel.bloecher@sap.com>

### DIFF
--- a/docs/adrs/001-event-based-feature-extraction.md
+++ b/docs/adrs/001-event-based-feature-extraction.md
@@ -1,0 +1,129 @@
+Architecture Decision Record 001
+
+---
+
+# Event-Based Feature Extraction
+
+## Background
+
+```mermaid
+graph TB
+subgraph Syncer[Syncer Service]
+direction TB
+SP1[OpenStack Objects]
+SP2[Prometheus Metrics]
+SP3[Other Data Sources]
+end
+DBS[(Synced Data in Database)]
+SP1 --> DBS
+SP2 --> DBS
+SP3 --> DBS
+subgraph FeatureExtractor[Feature Extractor Service]
+direction TB
+FE1[Contended Hostsystems]
+FE2[Best Placement for Flavors]
+FE3[Other Knowledge]
+end
+DBS --> FE1
+DBS --> FE2
+DBS --> FE3
+DBF[(Extracted Features in Database)]
+FE1 --> DBF
+FE2 --> DBF
+FE3 --> DBF
+Scheduler[Scheduler]
+DBF --> Scheduler
+```
+
+Currently we have an architecture as given in the diagram above:
+
+- The syncer service fetches data from remote sources (prometheus, openstack, in the future maybe kubernetes, ...) and inserts it into the database. Note that the synced data may be raw and unprocessed, such as prometheus timeline data.
+- The feature extractor reads this unprocessed data from the database and extracts features from it. Features provide a condensed, optimized knowledge for scheduling. For example, they can combine information from multiple data sources so that this information is not recalculated on each scheduler execution.
+- The scheduler uses this knowledge to make weighing/filtering decisions. Ideally, only the features relevant for this request (e.g. best placement for this flavor) are queried from the database, providing rapid scheduling decisions.
+
+Now, the more recent the provided features are, the better the scheduling decision can be made. For example, if a large vm flavor is placed on one of the hosts, this influences directly how future vms should be placed. In this case, the "best placement for flavors" feature (see above) needs to be recalculated as soon as possible.
+
+However, currently the feature extraction and syncing are run completely without knowledge of each other, in a 1-minute cycle. Thus, information that was synced may only propagate to the features after this cycle was restarted again. This may lead to inconsistencies i.e. to bad scheduling, and in the worst case, failed scheduling.
+
+## Proposed Solution
+
+Instead of executing the feature extraction periodically, the proposed solution would be to extract the features as soon as the underlying data has changed. This means that the syncer needs to communicate data changes to the feature extractor. Furthermore, the feature extractor needs to know which part of the dataset has changed, to not re-calculate all features. Furthermore, the update should propagate if [a feature depends on another feature](https://github.com/cobaltcore-dev/cortex/blob/d1698cfa7a07a1eafacfc56dd2545ee1e28da40b/helm/cortex/values.yaml#L269).
+
+Cons:
+- Added complexity to the syncer and the feature extractor
+- Syncer + feature extractor are no longer completely isolated except shared db usage
+- Need to consider mqtt triggers when implementing new extractors/syncers
+
+Pros:
+- Features that rarily change are not unnecessarily recalculated
+- Quicker response to changes in the underlying data, potentially less inconsistencies
+- Concept can be extended in the future, e.g. notifying the syncer when to fetch new data
+
+The following implementation options are available.
+
+### Implementation Option 1: Use Message Broker
+
+Add an message broker into the architecture and connect the syncer + feature extractor with it. Updates to data can be published on topics. Topics are defined in a consistent way, e.g. `triggers/sync/openstack/nova/types/hypervisors` when hypervisors change. All new syncer plugins and feature extractor plugins follow this schema.
+
+E.g., VerneMQT (MQTT) is a message broker that could be used.
+
+Cons:
+- Need to add mqtt broker and maintain that (operational overhead)
+
+Pros:
+- Existing service separation can stay for now, no refactoring of the structure needed
+- Decoupling & loose coupling, communicate via well-defined mqtt topics
+- Lightweight implementation
+
+### Implementation Option 2: Merge Syncer and Feature Extractor
+
+Merge the syncer and feature extractor service to one service (e.g. "", "")
+
+Cons:
+- Loss of service separation, we make it much harder to separate the services later on
+- Needs refactoring of the go modules, since they are separated by scheduler/syncer/features currently
+
+Pros:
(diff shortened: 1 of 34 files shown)


docs/adrs/001-event-based-feature-extraction.md
```diff
@@ -0,0 +1,126 @@
...
+- No support for future logic like "scheduler spawned VM" --> "monitor VM in OpenStack" --> "spawned" --> "resync"
+
+Pros:
+- No change effort
+- No additional services introduced into the architecture that need to be maintained
+- Current service isolation persists
+
+## Decision and Consequences
+
+Among the available options, option 1 was chosen. As an implementation variant for now, mqtt will be chosen. This also means that the mqtt broker will become mandatory. Option 4 should be conceptually investigated further while we move forward with an iterative, learning-by-doing approach.
```
mblos, line 126: It is not option 4, but option 3 which is considered long term.
Could also rephrase a bit

```suggestion
Among the available options, **Option 1: Message Broker** was chosen as the immediate solution. This approach allows us to move forward with minimal disruption to the current architecture while addressing the need for event-driven design. MQTT will be used as the message broker, making it a mandatory component of the system.

However, **Option 3: Generic Distribution Task Pipeline** is recognized as a more robust and scalable long-term solution. While it offers significant flexibility and automation, it is currently considered overkill due to its high complexity and the substantial effort required for implementation.
As the system evolves and the demand for more advanced task handling grows, we will re-evaluate the design (e.g., **Option 3**) when it becomes necessary.
```

internal/features/plugins/kvm/node_exporter_host_cpu_usage.go
```diff
@@ -34,6 +34,13 @@ func (*NodeExporterHostCPUUsageExtractor) GetName() string {
 	return "node_exporter_host_cpu_usage_extractor"
 }
 
+// Get message topics that trigger a re-execution of this extractor.
+func (NodeExporterHostCPUUsageExtractor) Triggers() []string {
+	return []string{
+		"triggers/sync/prometheus/alias/node_exporter_cpu_usage_pct",
```
mblos, line 40: What about having a central place with all the topics defined as constants?
..also for all other places

docs/adrs/001-event-based-feature-extraction.md
```diff
@@ -0,0 +1,126 @@
+Architecture Decision Record 001
+
+---
+
+# Event-Based Feature Extraction
+
+## Background
```
mblos, line 7: Would be nice to use the TAM schema for the diagram

internal/sync/openstack/placement_sync.go
```diff
@@ -22,15 +23,19 @@ type placementSyncer struct {
 	conf PlacementConf
 	// Placement API client to fetch the data.
 	api PlacementAPI
+	// MQTT client to publish mqtt data.
+	// Only initialized if mqtt is enabled.
```
mblos, line 27: Did I overlooked it.. there is no enabled flag anymore

internal/sync/openstack/nova_sync.go
```diff
@@ -73,23 +73,38 @@ func (s *novaSyncer) Init(ctx context.Context) {
...
-// Sync the OpenStack nova objects.
+// Sync the OpenStack nova objects and publish triggers if enabled.
 func (s *novaSyncer) Sync(ctx context.Context) error {
 	// Only sync the objects that are configured in the yaml conf.
 	if slices.Contains(s.conf.Types, "servers") {
 		if _, err := s.SyncServers(ctx); err != nil {
 			return err
 		}
+		// If enabled, notify others of the update.
+		if s.mqttClient != nil {
```
mblos, line 84: Would be better to have this check as part of the mqtt lib. E.g. why does a service check if there is a mqtt connection, but simply ignores if there is none. Does the service start up and starts fetching data from e.g. OpenStack also in the case when there is no mqtt connection?

internal/features/pipeline.go
```diff
@@ -90,30 +97,61 @@ func NewPipeline(config conf.FeaturesConfig, database db.DB, m Monitor) FeatureE
...
 
-	// Print out the execution order to the log.
-	slog.Info("feature extractor: dependency graph resolved")
-	for i, extractors := range executionOrder {
-		for _, extractor := range extractors {
-			slog.Info(
-				"feature extractor: execution order",
-				"group", i, "name", extractor.GetName(),
-			)
+	// Resolve which feature extractors to execute when triggers are received.
```
mblos, line 100: Personal opinion - I would prefer to have smaller functions .. which makes code easier to read and more testable.

docs/adrs/001-event-based-feature-extraction.md
```diff
@@ -0,0 +1,126 @@
...
+
+```mermaid
+graph TB
+subgraph Syncer[Syncer Service]
```
mblos, line 11: Out of scope for this PR.. but would be nice to rename the services (discuss offline)

mblos (dismissed):
partial review

internal/features/plugins/kvm/node_exporter_host_cpu_usage.go
PhilippMatthes: Hmm, maybe something like sync.TriggerXYZ and feature.TriggerXYZ contained in a `triggers.go` file? Let me hack this in.

docs/adrs/001-event-based-feature-extraction.md
PhilippMatthes: Sure, maybe this would be a good follow-up for this PR? I liked the idea that github can display mermaid diagrams out of the box, thus we don't have to include images. But im open to your idea

internal/features/plugins/kvm/node_exporter_host_cpu_usage.go
PhilippMatthes: See https://github.com/cobaltcore-dev/cortex/pull/67/commits/d46c185a6c1890eb7063dd3721038b1e44f5c0b3

internal/sync/openstack/nova_sync.go
PhilippMatthes: Good catch, this check is no longer necessary. `s.mqttClient` here is an interface, which means it can also be `nil`. It was nil when the mqtt client was not enabled. However, since it is mandatory now, we can disable this check. I just need to implement a mock mqtt client for the unit tests.

PhilippMatthes: Implemented in https://github.com/cobaltcore-dev/cortex/pull/67/commits/6eec8f115e95f9b9f86ffbb15b617539f2ee5f89

docs/adrs/001-event-based-feature-extraction.md
PhilippMatthes: Happy to discuss

internal/features/pipeline.go
PhilippMatthes: Yep, this function is definitely bloated. So far I didn't have enough pain to refactor this, but now that we are two, let me polish this.

PhilippMatthes: Refactored in https://github.com/cobaltcore-dev/cortex/pull/67/commits/f74b9aad432b4a1d7b11284c4005fbdfc6e0e2af

Working on tests.

PhilippMatthes: Added tests in https://github.com/cobaltcore-dev/cortex/pull/67/commits/4f1b7f2ad84a2900608fdbf42775e75b3f75d8e3

internal/sync/openstack/placement_sync.go
PhilippMatthes: Thanks, good catch, removed these occurences in fd21577b1345aa5b02206b7b6b3b5c8741a26ba5

internal/conf/graph_test.go
```diff
@@ -70,3 +70,102 @@ func TestDependencyGraph_Resolve(t *testing.T) {
 		})
 	}
 }
+
+func TestDependencyGraph_DistinctSubgraphs(t *testing.T) {
```
mblos, line 74: I miss a test with cyclic dependencies. Would be good to have one here

testlib/mqtt/mock.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2025 SAP SE
+// SPDX-License-Identifier: Apache-2.0
+
+package mqtt
+
+import (
+	pahomqtt "github.com/eclipse/paho.mqtt.golang"
+)
+
+// Mock mqtt client that does nothing and can be used for testing.
```
mblos, line 10: Good place to add more functionality here.. e.g. actually sending messages such that we can have integration test cases. But out of scope for this PR.

mblos (approved):

internal/conf/graph_test.go
PhilippMatthes: Added in 67a0ccaec944269cd4336d106621ba3501dafdbb

mblos (approved):

PR #68, opened 2025-03-20
Fix crash resulting from large number of servers returned from Nova

PhilippMatthes:


auhlig (approved):

internal/sync/openstack/nova_api.go
```diff
@@ -77,7 +77,12 @@ func (api *novaAPI) GetAllServers(ctx context.Context, changedSince time.Time) (
 			timer := prometheus.NewTimer(hist)
 			defer timer.ObserveDuration()
 		}
-		lo := servers.ListOpts{AllTenants: true, ChangesSince: changedSince.Format(time.RFC3339)}
+		// It is important to omit the changes-since parameter if it is nil.
+		// Otherwise Nova will return huge amounts of data since the beginning of time.
+		lo := servers.ListOpts{AllTenants: true}
+		if changedSince != nil {
+			lo.ChangesSince = changedSince.Format(time.RFC3339)
```
mblos, line 84: Why not adding a test case that checks exactly for this? Same for `GetAllFlavors`

mblos (dismissed):

PR #76, opened 2025-03-25
Fallback to original hosts if all are filtered out

PhilippMatthes:
See also: https://github.com/sapcc/nova/pull/535

This adds more safety, in case some scheduler step decides to filter out all hosts from the Nova request. For now we consider this case a bug, and if we are confident enough, we can remove this check and let the scheduling fail on purpose.


PhilippMatthes:
We decided not to support this but instead implement alerts that catch if something is wrong. The rest will be handled over playbooks (how to scale down cortex, etc.).

PR #83, opened 2025-03-31
Validate same host number in and out for weighers

PhilippMatthes:
- Add step validator as wrapper, similar to step monitor
- Move general logging out of step monitor
- Enable all validations by default, but can be disabled through config
- No generics in step monitor

### DIFF
--- a/internal/scheduler/validation.go
+++ b/internal/scheduler/validation.go
@@ -19,21 +19,21 @@ import (
 type disabledValidations = conf.SchedulerStepDisabledValidationsConfig
 
 // Wrapper for scheduler steps that validates them before/after execution.
-type StepValidator[S plugins.Step] struct {
+type StepValidator struct {
 	// The wrapped step to validate.
-	Step S
+	Step plugins.Step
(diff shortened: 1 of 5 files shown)


internal/scheduler/validation.go
```diff
@@ -0,0 +1,65 @@
...
+
+// The config type has a long name, so we use a shorter alias here.
+// The name is intentionally long to make it explicit that we disable
+// validations for the scheduler step instead of enabling them.
+type disabledValidations = conf.SchedulerStepDisabledValidationsConfig
+
+// Wrapper for scheduler steps that validates them before/after execution.
+type StepValidator[S plugins.Step] struct {
+	// The wrapped step to validate.
+	Step S
```
felix-kaestner, line 24:
```suggestion
	Step plugins.Step
```
What does the generic bring as value?

PhilippMatthes: Actually none here. Will adjust this

auhlig (approved):

PR #108, opened 2025-04-25
Improve scheduler KPIs for host reorderings

PhilippMatthes:
This pr adds KPIs to not only measure how many hosts were re-ordered in the scheduling pipeline, but also to measure how high the impact of this re-ordering was.

### DIFF
--- a/internal/scheduler/monitor.go
+++ b/internal/scheduler/monitor.go
@@ -26,7 +26,7 @@ type Monitor struct {
 	stepHostWeight *prometheus.GaugeVec
 	// A histogram to observe how many hosts are removed from the state.
 	stepRemovedHostsObserver *prometheus.HistogramVec
-	// Observer for the number of reorderings conducted by the scheduler, by step.
+	// Histogram measuring where the host at a given index came from originally.
 	stepReorderingsObserver *prometheus.HistogramVec
 	// A histogram to measure how long the pipeline takes to run in total.
 	pipelineRunTimer prometheus.Histogram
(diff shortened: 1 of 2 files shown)


internal/scheduler/levenshtein_test.go
```diff
@@ -0,0 +1,196 @@
...
+			expected: 2, // Host shifted by 2 positions
+		},
+		{
+			name: "Substitution pair",
+			a:    []string{"host1", "host2", "host3"},
+			edits: []LevEdit{
+				LevSub("host3", 0),
+				LevSub("host1", 2),
+			},
+			expected: 4, // Swapped distance counted twice
```
anokfireball, line 164: My fluency in the Levenshtein algorithm is pretty rusty.
Is the transformation [1,2,3] -> [3,2,1] determined to be swap only rather than 2 shifts [1,2,3] -> [3,1,2] -> [3,2,1] because that transform nominally consists of fewer transformations? 2 substitutions vs. 2 deletions + 2 insertions.

Asking because - if I understand the hostShift algo correctly - the 2 swaps would have a shift value of 4 while the 2 deletions + 2 insertions would have a shift value of 3.

PhilippMatthes: But I definitely need to check if levenshtein calculates the insertions + deletions on the static string, meaning that the value would be 4 contrasting your example where you assume transitive edits on the string: 
[a, b, c] to [c, b, a] -> del a[0] ins a[2], del c[2] ins c[0] -> shift 4

PhilippMatthes: After reconsidering, there may be more cases where the shift metric would be broken, e.g.
```
levenshtein(
  [host1 host2 host3 host4 host5 host6], 
  [host4 host1 host2 host3 host6 host5]
) 
edits = [{insert host4 0} {sub host6 4} {delete host6 5}]
```
So I need to come up with something different.

PR #125, opened 2025-05-09
Fail hard on disconnects from mqtt and postgres

PhilippMatthes:
This should avoid any services dangling around without an active connection to either.


PhilippMatthes:
Need to fix the mqtt tests.

PR #131, opened 2025-05-16
Enhance MQTT reconnection logic

SoWieMarkus:
- Added configurable retry logic
- Config Max Retries attempts, Initial Retry Delay, Retry Interval
- Added test to simulate connection loss

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -24,19 +24,39 @@ helm_repo(
     labels=['Repositories'],
 )
 
-# Build the helm charts
-local('test -f ./helm/cortex/Chart.lock || helm dep up ./helm/cortex')
-local('test -f ./helm/mqtt/Chart.lock || helm dep up ./helm/mqtt')
-local('test -f ./helm/prometheus/Chart.lock || helm dep up ./helm/prometheus')
-local('test -f ./helm/prometheus-operator/Chart.lock || helm dep up ./helm/prometheus-operator')
-local('test -f ./helm/postgres/Chart.lock || helm dep up ./helm/postgres')
+def synced_helm(path_to_chart, name, values=[]):
+    """
+    Tilt provides a helm() function that renders a chart and watches it.
+    However, it also declares file watchers for all files in the chart.
+    This means that if we want to run helm dep build to sync the chart/
+    dependencies before helm(), this will end in an infinite deployment loop.
+    Therefore, we need to selectively watch the files we care about.
+    """
+
+    # Build the chart dependencies from the Chart.lock.
+    local('helm dep build ' + path_to_chart)
+
+    # Build the command to get the rendered kubernetes yaml.
+    cmd = 'helm template'
+    for value_file in values:
+        cmd += ' -f ' + value_file
+    cmd += ' --name-template ' + name
+    cmd += ' ' + path_to_chart
+
+    watch_file(path_to_chart + '/Chart.yaml')
+    watch_file(path_to_chart + '/Chart.lock')
+    for value_file in values:
+        watch_file(value_file)
+    watch_file(path_to_chart + '/templates')
+
+    return local(cmd)
 
 ########### Cortex Core Services
 tilt_values = os.getenv('TILT_VALUES_PATH')
 docker_build('ghcr.io/cobaltcore-dev/cortex', '.', only=[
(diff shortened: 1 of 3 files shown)


internal/mqtt/mqtt.go
```diff
@@ -70,7 +106,7 @@ func (t *client) Connect() error {
 
 	client := mqtt.NewClient(opts)
 	if conn := client.Connect(); conn.Wait() && conn.Error() != nil {
-		panic(conn.Error())
+		return conn.Error()
```
PhilippMatthes, line 109: Currently we lazily connect the mqtt client on the first `.Publish` call. We should change that to connecting the mqtt broker on the service startup. Because now, if the initial connection fails, this may leave our service dangling around if `onUnexpectedConnectionLoss` is not called. We should then also remove the lazy connection in the `Publish` function

internal/mqtt/mqtt.go
```diff
@@ -127,6 +175,11 @@ func (t *client) Subscribe(topic string, callback mqtt.MessageHandler) error {
 		return token.Error()
 	}
 	slog.Info("subscribed to topic", "topic", topic)
+
+	if t.subscriptions == nil {
+		t.subscriptions = make(map[string]mqtt.MessageHandler)
+	}
+	t.subscriptions[topic] = callback
```
PhilippMatthes, line 182: Not sure about the nil check here, maybe fix the tests instead if they break.

internal/mqtt/mqtt_test.go
```diff
@@ -68,7 +69,66 @@ func TestSubscribe(t *testing.T) {
...
+	// Wait for the client to reconnect and receive the message
+	select {
+	case <-reconnected:
+		t.Log("client successfully reconnected and received message")
+	case <-time.After(10 * time.Second):
+		t.Fatal("client did not reconnect and receive message in time")
+	}
+
 	c.Disconnect()
+	otherClient.Disconnect()
```
PhilippMatthes, line 133: Like this test.

internal/mqtt/mqtt.go
```diff
@@ -28,18 +29,49 @@ type client struct {
...
+
+	for retry := range t.conf.ConnectionLostMaxRetries {
+		slog.Info("attempting to reconnect to mqtt broker", "attempt", retry+1, "url", t.conf.URL)
+
+		if err := t.Connect(); err != nil {
+			slog.Error("failed to reconnect to mqtt broker", "err", err)
+			if retry < t.conf.ConnectionLostMaxRetries-1 {
+				time.Sleep(time.Duration(t.conf.ConnectionLostRetryInterval) * time.Second)
+			}
+			t.client = nil
```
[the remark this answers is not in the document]
SoWieMarkus: Yes. Because if the client is != nil the Connect() call immediately returns in the next try https://github.com/cobaltcore-dev/cortex/blob/9f6c624bfe0401310e8893f24b771ae9aabe45fa/internal/mqtt/mqtt.go#L80

internal/db/db.go
```diff
@@ -77,21 +79,26 @@ func NewPostgresDB(c conf.DBConfig, registry *monitoring.Registry) DB {
...
 	}
-	return DB{DbMap: dbMap}
+	return DB{DbMap: dbMap, monitor: monitor}
 }
 
 // Check periodically if the database is alive. If not, panic.
 func (d *DB) CheckLivenessPeriodically() {
 	var failures int
 	for {
+		maxRetries := 20
```
PhilippMatthes, line 89: Make this configurable as well?

internal/mqtt/mqtt.go
```diff
@@ -28,18 +29,52 @@ type client struct {
 	client *mqtt.Client
 	// Lock to prevent concurrent writes to the MQTT client.
 	lock *sync.Mutex
+	// monitor for mqtt related metrics
+	monitor Monitor
+	// all current subscribed topics of the client
```
PhilippMatthes, line 34: Nit: Uppercase comments maybe to stay consistent?

PhilippMatthes (approved):
Very nice! Just one or two small things then send it 🚀

PhilippMatthes (approved):

PhilippMatthes (approved):

PR #140, opened 2025-05-23
Properly name chart dirs and add maintainers

PhilippMatthes:


helm/cortex-mqtt/charts/.gitignore
SoWieMarkus, line 1: I noticed that the old folders with ignored files were not deleted, but the .gitignore file was moved to the new directory. Which led to git not ignoring the content of the old charts/ directories. This probably isn’t a big issue since only two of us are working on this, but I think it would make sense to add a .gitignore in the helm/ directory that handles all the ignored files inside the charts/. That way, we can also avoid the error where Helm can’t handle the .gitignore file.

SoWieMarkus (commented):
What do you think about the comment? Also tilt isnt building for me right now because of the prometheus-operator. I have a quick look.

helm/cortex-mqtt/charts/.gitignore
SoWieMarkus: https://github.com/cobaltcore-dev/cortex/pull/140/commits/cc3f3bdeb344909d78fc2baacffc0015ee12c481

PR #143, opened 2025-05-27
Avoid unnecessary burst calculation of features

SoWieMarkus:
- Added a `recencySeconds` config option so you can control how much time needs to pass before a feature extractor runs again.
- Implemented logic to check if the last update is at least `recencySeconds` ago before extracting features.
- Adjusted tests and code to work with the new recency logic.

@PhilippMatthes after your vacation I’d like to discuss with you what a suitable recency would be for each feature extractor.

### DIFF
--- a/internal/extractor/pipeline.go
+++ b/internal/extractor/pipeline.go
@@ -195,10 +195,15 @@ func (p *FeatureExtractorPipeline) extract(order [][]plugins.FeatureExtractor) {
 			wg.Add(1)
 			go func(extractor plugins.FeatureExtractor) {
 				defer wg.Done()
+				if !extractor.NeedsUpdate() {
+					slog.Info("feature extractor: skipping extraction", "extractor", extractor.GetName())
+					return
+				}
 				if _, err := extractor.Extract(); err != nil {
 					slog.Error("feature extractor: failed to extract features", "error", err)
 					return
 				}
+				extractor.MarkAsUpdated()
 			}(extractor)
 		}
 		wg.Wait()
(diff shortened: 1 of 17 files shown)


internal/extractor/pipeline.go
```diff
@@ -195,10 +195,15 @@ func (p *FeatureExtractorPipeline) extract(order [][]plugins.FeatureExtractor) {
 			wg.Add(1)
 			go func(extractor plugins.FeatureExtractor) {
```
PhilippMatthes, line 196: Let's clean this up: `go func () {...}`

internal/extractor/plugins/base.go
```diff
@@ -52,3 +59,23 @@ func (e *BaseExtractor[Opts, F]) Extracted(fs []F) ([]Feature, error) {
...
+	if e.RecencySeconds <= 0 {
+		return true
+	}
+	return time.Since(*e.UpdatedAt) > time.Duration(e.RecencySeconds)*time.Second
+}
+
+// Mark the extractor as updated by setting the UpdatedAt field to the current time.
+func (e *BaseExtractor[Opts, F]) MarkAsUpdated() {
+	e.UpdatedAt = new(time.Time)
+	*e.UpdatedAt = time.Now()
```
PhilippMatthes, line 80: Doesn't this work?
```suggestion
  now := time.Now()
	e.UpdatedAt = &now
```

PhilippMatthes (approved):
Really nice work! Only 2 nits

PR #150, opened 2025-06-04
Initial Manila passthrough to build API

PhilippMatthes:


Tiltfile
```diff
@@ -42,13 +42,19 @@ k8s_resource('cortex-extractor', port_forwards=[
...
+    port_forward(8003, 8080),
+    port_forward(8004, 2112),
+], links=[
+    link('localhost:8003/metrics', '/metrics'),
+], labels=['Core-Services'])
+k8s_resource('cortex-scheduler-manila', port_forwards=[
+    port_forward(8005, 8080),
+    port_forward(8006, 2112),
 ], links=[
     link('localhost:8003/metrics', '/metrics'),
```
SoWieMarkus, line 54: Should this be localhost:8006?

Tiltfile
```diff
@@ -42,13 +42,19 @@ k8s_resource('cortex-extractor', port_forwards=[
...
 ], labels=['Core-Services'])
 k8s_resource('cortex-kpis', port_forwards=[
-    port_forward(8004, 2112),
+    port_forward(8007, 2112),
 ], links=[
     link('localhost:8004/metrics', '/metrics'),
```
SoWieMarkus, line 59: Should this be localhost:8007?

internal/scheduler/manila/plugins/result.go
```diff
@@ -0,0 +1,30 @@
...
+	//	{
+	//	  "max cpu contention": {
+	//	     "unit": "cpu contention [%]",
+	//	     "hosts": { "share host 1": 10, "share host 2": 10 }
+	//	   },
+	//     ...
+	//	}
+	//
+	// These statistics are used to display the step's effect on the hosts.
+	// For example: max cpu contention: before [ 100%, 50%, 40% ], after [ 40%, 50%, 100% ]
```
SoWieMarkus, line 21: Is that a "good" example for manila scheduler step results?

SoWieMarkus (approved):

Tiltfile
PhilippMatthes: See https://docs.tilt.dev/api.html#api.port_forward

internal/scheduler/manila/plugins/result.go
PhilippMatthes: I think so.

PR #151, opened 2025-06-05
Support trait + hypervisor type scoping and resource balancing

PhilippMatthes:


internal/extractor/plugins/shared/host_space_test.go
```diff
@@ -0,0 +1,109 @@
...
+			DiskLeftGB:   400,
+			DiskLeftPct:  50.0,
+		},
+	}
+
+	for i, space := range spaces {
+		if space != expected[i] {
+			t.Errorf("expected %+v, got %+v", expected[i], space)
+		}
+	}
```
SoWieMarkus, line 108: You could check that `ironic` hypervisors are filtered out.

SoWieMarkus (approved):

PR #152, opened 2025-06-09
PoC: Introduce read-only postgres replicas for accelerated scheduling queries

PhilippMatthes:
This is a proof-of-concept of how we can decrease the query time in the scheduler service further by introducing 2 read-only replicas for the postgres instance. This is a breaking change in the postgres chart, meaning we need to either migrate the dbs by hand or reset them. 

In preliminary testing this reduced query time in the scheduler by ~80%, caveat this was no reproducable clean measurement.


PhilippMatthes:
No longer applicable since we removed the bitnami charts. I'm also going to delete the branch.

PR #164, opened 2025-06-13
Refactored host_utilization KPI to use host_space Feature

SoWieMarkus:
~Refactored the `host_utilization` KPI to use the `feature_host_space` table as the primary data source, since both previously implemented similar logic.~

~Because the `feature_host_space` table does not contain the hypervisor's `id` and `hostname`—which are used as labels in the Prometheus metrics—I added a join with the `openstack_hypervisor` table to retrieve this information.
If these labels are not essential for this KPI, the join could be omitted for a simpler implementation. Whats your opinion @PhilippMatthes~

- Rewrite `host_space_extractor` to `host_utilization_extractor` to consider overcommit (wasn't possible with the host space extractor)
- Use new extractor to export host utilization kpis properly
- Removed `flavor bin packer` since it provides the same functionality as `resource balancer`
- Refactor nova object sync: don't upsert hypervisors on id, rename sync function


SoWieMarkus:
I'll review the changes to see if the logic is correct. The Plutono chart appears different than before.

PhilippMatthes (approved):

PR #166, opened 2025-06-16
Added nova aggregates syncer

SoWieMarkus:
## Changes  
- Added Nova Aggregate Syncer
- Added Availability Zone as label to `host_utilization_kpi`

## Reverted Changes

We don't want to build a custom frontend already. We stick with plutono dashboards for now. Thats why the following changes have been reverted:

- Added new cortex core service `visualizer-api`
- Added endpoint that atm just returns the `host-utilization` table (will be adjusted to calculate stacked bar chart data)
- Reduced requested memory for pods to 512Mi

### DIFF
--- a/helm/cortex/values.yaml
+++ b/helm/cortex/values.yaml
@@ -404,97 +404,139 @@ conf:
       # the order will be maintained when the weights are applied to the
       # input Nova weights.
       plugins:
-        # VMware specific scheduler steps.
+        # Avoid co-location of noisy projects on the same host.
+        # This step depends on vrops metrics and is thus only
+        # executed for VMware vms.
         - name: vmware_anti_affinity_noisy_projects
+          scope:
+            specSelectors:
+              - subject: vmware
+                type: bool
+                value: true
+                action: continue
           options:
-            # Min-max scaling for avg CPU usage of the project on a host.
             avgCPUUsageLowerBound: 0 # pct
             avgCPUUsageUpperBound: 10 # pct
             avgCPUUsageActivationLowerBound: 0.0
             avgCPUUsageActivationUpperBound: -1.0
           dependencies:
-            extractors:
-              - vrops_project_noisiness_extractor
-        - name: vmware_avoid_long_term_contended_hosts
-          options:
-            # Min-max scaling for avg CPU contention on the host.
-            avgCPUContentionLowerBound: 0 # pct
-            avgCPUContentionUpperBound: 10 # pct
-            avgCPUContentionActivationLowerBound: 0.0
-            avgCPUContentionActivationUpperBound: -0.75
-            # Min-max scaling for max CPU contention on the host.
-            maxCPUContentionLowerBound: 0 # pct
-            maxCPUContentionUpperBound: 10 # pct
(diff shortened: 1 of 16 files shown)


internal/sync/openstack/nova/nova_sync.go
```diff
@@ -108,6 +111,15 @@ func (s *NovaSyncer) Sync(ctx context.Context) error {
 			go s.MqttClient.Publish(TriggerNovaMigrationsSynced, "")
 		}
 	}
+	if slices.Contains(s.Conf.Types, "aggregates") {
+		changedAggregates, err := s.SyncChangesAggregates(ctx)
```
PhilippMatthes, line 115: There is a typo. `SyncAllAggregates` better?

internal/sync/openstack/nova/nova_sync.go
```diff
@@ -260,3 +272,21 @@ func (s *NovaSyncer) SyncChangedMigrations(ctx context.Context) ([]Migration, er
...
 	return changedMigrations, nil
 }
+
+func (s *NovaSyncer) SyncChangesAggregates(ctx context.Context) ([]Aggregate, error) {
+	tableName := Aggregate{}.TableName()
+	lastSyncTime := s.getLastSyncTime(tableName)
+
+	defer s.setLastSyncTime(tableName, time.Now())
+
+	// TODO can't pass timestamp to the nova api, so we just fetch all aggregates.
```
PhilippMatthes, line 282: TODO? This won't change or no?

internal/sync/openstack/nova/nova_types.go
```diff
@@ -246,3 +246,17 @@ func (Migration) TableName() string { return "openstack_migrations" }
 
 // Index for the openstack model.
 func (Migration) Indexes() []db.Index { return nil }
+
+type Aggregate struct {
+	UUID             string  `json:"uuid" db:"uuid,primarykey"`
+	Name             string  `json:"name" db:"name,primarykey"`
+	AvailabilityZone *string `json:"availability_zone" db:"availability_zone"`
+	ComputeHost      string  `json:"compute_host" db:"compute_host,primarykey"` // Host that the aggregate is associated with.
```
PhilippMatthes, line 254: Particular reason why we make a triple primary key here? Shouldn't UUID be sufficient?

PhilippMatthes (approved):
Very cool! A bit polishing but then, lgtm

internal/sync/openstack/nova/nova_types.go
SoWieMarkus: No. Each aggregate has multiple hosts and we create one entry for each host.

| UUID | Name   | Host  | Properties |
|------|--------|-------|------------|
| 1    | ger01  | host1 | {}         |
| 2   | else01  | host1 | {}         |
| 1    | ger01  | host2 | {}         |
| 3   | ger02   | host3 | {}         |

`SELECT * FROM openstack_aggregates ORDER BY compute_host;` if you want to have a look. [Docs](https://docs.openstack.org/api-ref/compute/#list-aggregates)

PhilippMatthes: Ah, you're right.

PR #174, opened 2025-06-20
Added cpu model as label for the cortex_host_utilization_per_host_pct

SoWieMarkus:
- Adjusted tests to check if the join inside the kpi collect method works correctly
- Added cpu_model as label


PhilippMatthes (approved):
Cool!

Let's see how long the feature extraction takes in prod.

PR #188, opened 2025-07-03
Add alerts for high CPU, memory, and disk utilization on hosts

SoWieMarkus:
Added alerts if cortex detects resource usage above 100% for more than 5m for `cpu`, `memory` and `disk`


PhilippMatthes (approved):
Cool! Maybe some wording changes as discussed offline, and a bit more description would be nice

PR #196, opened 2025-07-09
Add extractor for mapping hypervisors to projects and domains

SoWieMarkus:
> [!NOTE]
> This implementation does not work with SQLite. See [example SQLite-compatible implementation](https://github.com/cobaltcore-dev/cortex/pull/196/commits/36f0e1948f6be98b9d19895086a7ad93e8d0bde6) for reference.

Added new feature extractor that stores all domains and projects per compute host as comma-separated values.

| compute_host | project_names      | project_ids | domain_names      | domain_ids |
|--------------|-------------------|-------------|-------------------|------------|
| host1        | project1,project2 | p1,p2       | domain1,domain2   | d1,d2      |
| host2        | project2          | p2          | domain2           | d2         |


SoWieMarkus:
But I mean it's not just about the test. The whole feature extractor wouldn't work with sqlite. Didn't we try to make them compatible with both Postgres and SQLite?

PhilippMatthes (approved):
Very cool!

We should check how long the kpis take to scrape in the biggest regions, because there is the potential for timeouts to happen if we take too long with our metrics logic.

PR #197, opened 2025-07-09
Fix replay command and add manila visualizer

PhilippMatthes:


SoWieMarkus (approved):
Didn't test it myself yet but code lgtm

PR #199, opened 2025-07-10
Added host domains and projects as label to host_utilization api

SoWieMarkus:


PhilippMatthes (approved):
Very nice!
Buzzing to see this in action

PR #203, opened 2025-07-14
PoC: Configure features through CRD

PhilippMatthes:
Keeping this open for now so that we have a reference


PhilippMatthes:
Better example https://github.com/cobaltcore-dev/cortex/pull/249

PR #208, opened 2025-07-16
Release cortex chart v0.23.3 and cortex-prometheus chart v0.4.2

PhilippMatthes:
- Add alert for failed deschedulings
- Fix multiple aliases for same step overriding config (leading to only one of hana binpacking or resource balancing being initialized)


PhilippMatthes:
Need to bump the cortex chart once the CI is done

SoWieMarkus (approved):

mblos (approved):

PR #230, opened 2025-08-05
Add host details feature for capacity dashboard

SoWieMarkus:
- Added new feature extractor `sap_host_details_extractor` which maps the following attributes to each host:
  - `cpu_architecture` (cascade_lake, sapphire_rapids)
  - `hypervisor_family` (vmware, kvm)
  - `workload_type` (general-purpose, hana)
  - `availability_zone`
  - `enabled` (Check for traits that will disable this hypervisor)
- Refactor `host_utilization_kpi` to use `sap_host_details_extractor` for the labels
- Moved `host_total_capacity_kpi` to own file

### DIFF
--- a/plutono/provisioning/dashboards/cortex-capacity.json
+++ b/plutono/provisioning/dashboards/cortex-capacity.json
@@ -1744,7 +1744,7 @@
           {
             "matcher": {
               "id": "byName",
-              "options": "Availble CPUs"
+              "options": "Available CPUs"
             },
             "properties": [
               {
@@ -1910,7 +1910,7 @@
         {
           "id": "calculateField",
           "options": {
-            "alias": "Availble CPUs",
+            "alias": "Available CPUs",
             "binary": {
               "left": "Value #TotalCPUQuery",
               "operator": "*",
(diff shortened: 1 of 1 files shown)


helm/library/cortex-core/Chart.yaml
```diff
@@ -6,4 +6,4 @@ name: cortex-core
 description: A Helm chart deploying Cortex core services.
 type: application
 version: 0.24.4
-appVersion: "sha-cc4a7b7"
+appVersion: "sha-111bf5d"
```
PhilippMatthes, line 9: I think this can be reverted.

internal/kpis/pipeline.go
```diff
@@ -25,10 +25,12 @@ var SupportedKPIs = []plugins.KPI{
 	// NetApp kpis.
 	&netapp.NetAppStoragePoolCPUUsageKPI{},
 	// Shared kpis.
+	&shared.HostTotalCapacityKPI{},
```
PhilippMatthes, line 28: Move this to "sap-specific kpis" and the associated plugins to a new folder?

PhilippMatthes (approved):
Very nice!

internal/kpis/pipeline.go
SoWieMarkus: Will do the same for the `running_vms` and `total_capacity` kpi since they also depend on the `sap_host_details` extractor

PR #242, opened 2025-08-13
Added host details check for status disabled and trait COMPUTE_STATUS_DISABLED

SoWieMarkus:
- Disable hypervisor with `openstack_hypervisor.status != 'enabled'`
- Disable hypervisor with trait `COMPUTE_STATUS_DISABLED`
- Added left join to also include hosts without a resource provider (is this even possible, just noticed while adjusting the test?)


SoWieMarkus:
Maybe add the `service_disabled_reason` field as disabled_reason

PhilippMatthes:
Just leaving this here. The feature also needs renaming i guess :)

PhilippMatthes (approved):
Looked at it today morning already, lgtm :)

PR #249, opened 2025-08-15
Schedule commitments with new reservations operator & crd

PhilippMatthes:
### Reservations operator

- Added syncer in operator that fetches commitments from limes and creates `ComputeReservation` resources for them. They are distinguished by `instance` or `bare` reservations.
- Added reconciler in operator that takes all `instance` reservations and tries to place them on a host using the initial placement pipeline. If the placement succeeds, the reservation is considered `active`.
- Added conf for supported hypervisor types, as we don't support the full filtering pipeline for vmware and ironic yet. `instance` reservations that don't match the hypervisor type are put into a terminal `failed` state.
- Added logic that puts all `bare` reservations into a terminal `failed` state as we don't know yet how to handle these reservations and generate initial placement requests for them.
- Added scaffolding with kubebuilder and separate dockerfile for operator, so we can reuse as many kubebuilder features as possible.

### Cortex core

- Added kubernetes client to capacity filter so that reservations are considered. Note that the filter is currently sandboxed meaning it will have no effect for now until we de-sandbox it.
- Added service account that binds to the appropriate reservations operator role, granting cortex core access to the reservations resources.
- Refactored config loading so that we can reuse it in the reservations operator.
- Added reservations operator api binds in go.mod so that cortex-core can access the v1alpha1 structs.

### CI/CD

- Added image build workflow for cortex reservations operator
- Added chart build workflow for cortex reservations operator
- Run unit tests of reservations module and report coverage

### Tilt setup

- Added reservations operator to the tilt setup.


PhilippMatthes:
Needs some more polishing but is otherwise ready for review.

internal/reservations/commitments.go
```diff
@@ -0,0 +1,352 @@
...
+
+	// The openstack project ID this commitment is for.
+	ProjectID string `json:"project_id"`
+	// The openstack domain ID this commitment is for.
+	DomainID string `json:"domain_id"`
+}
+
+// OpenStack flavor model as returned by the Nova API under /flavors/detail.
+// See: https://docs.openstack.org/api-ref/compute/#list-flavors
+type Flavor struct {
```
SoWieMarkus, line 88: Just out of curiosity: Why not reuse the nova.Flavor here?

internal/scheduler/nova/plugins/shared/filter_has_enough_capacity.go
```diff
@@ -4,15 +4,66 @@
...
 type FilterHasEnoughCapacity struct {
-	scheduler.BaseStep[api.ExternalSchedulerRequest, scheduler.EmptyStepOpts]
+	scheduler.BaseStep[api.ExternalSchedulerRequest, FilterHasEnoughCapacityOpts]
+
+	// Kubernetes client.
+	Client client.Client
+}
+
+func (s *FilterHasEnoughCapacity) Init(alias string, db db.DB, opts conf.RawOpts) error {
+	s.BaseStep.Init(alias, db, opts)
```
SoWieMarkus, line 43:
```suggestion
	if err := s.BaseStep.Init(alias, db, opts); err != nil {
		return err
	}
```

SoWieMarkus (approved):

internal/reservations/commitments.go
PhilippMatthes: I want to minimize the dependencies between packages here.

internal/reservations/api/v1alpha1/reservation.go
```diff
@@ -0,0 +1,125 @@
...
+
+// ReservationList contains a list of Reservation.
+type ReservationList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Reservation `json:"items"`
+}
+
+// Conform to the runtime.Object interface.
+func (in *ReservationList) DeepCopyObject() runtime.Object {
```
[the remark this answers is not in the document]
PhilippMatthes: Looking into it :)

SoWieMarkus (approved):

PhilippMatthes:
Integrating with kubebuilder. Needs major changes to the repository structure.

PhilippMatthes:
Kubebuilder is in -- I just need to polish everything once more.

PR #264, opened 2025-09-01
Removed project and domain labels from infrastructure kpis

SoWieMarkus:


SoWieMarkus:
Waiting for the next dashboard sync. If they still consider this filter useful I will partially revert these changes to add the labels back to the metrics. 

If they say we don't need this filter I think I will just delete the whole feature in general.

PR #292, opened 2025-09-15
Fix queries of commitments dashboard

SoWieMarkus:
Previously, when filtering by availability zones, the total number across all availability zones didn't add up correctly.

Using `max by (resource_name)` alone is effective for deduplication, but when the same `resource_name` exists across multiple availability zones, it only considers the maximum value across all zones rather than summing the individual zone values.

**Example of the issue:**

| Availability Zone | Resource Name | Units |
|-------------------|---------------|-------|
| AZ1               | Instance1     | 300   |
| AZ2               | Instance1     | 400   |

With the old query, only `400` units would be reported (the max), not the expected total of `700`.

### Solution
The new approach implements a two-step aggregation:

1. **Deduplicate entries** by finding the max value per `resource_name` and `availability_zone`
2. **Sum the results** by `resource_name` to get the correct total across zones

**Example with duplicates:**

| Availability Zone | Resource Name | Units | |
|-------------------|---------------|-------|-|
| AZ1               | Instance1     | 300   | |
| AZ1               | Instance1     | 301   | (duplicate) |
| AZ2               | Instance1     | 400   | |

-> **`max by (resource_name, availability_zone)`:**

| Availability Zone | Resource Name | Units |
|-------------------|---------------|-------|
| AZ1               | Instance1     | 301   |
| AZ2               | Instance1     | 400   |

-> **`sum by (resource_name)`:**
Total: `701` units ✓

### DIFF
--- a/plutono/provisioning/dashboards/cortex-commitments.json
+++ b/plutono/provisioning/dashboards/cortex-commitments.json
@@ -15,8 +15,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 1,
-  "iteration": 1754059701443,
+  "iteration": 1757917531460,
   "links": [],
   "panels": [
     {
@@ -26,7 +25,7 @@
         "h": 1,
         "w": 24,
         "x": 0,
-        "y": 0
+        "y": 8
       },
       "id": 7,
       "panels": [],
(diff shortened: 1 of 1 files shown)


PhilippMatthes (approved):
That's a proper + cool fix!

PR #297, opened 2025-09-17
Add extractor for resource utilization and resource commitments of a project

SoWieMarkus:
- add two new feature extractor
- one extracts the resource utilization of a project
- the other extracts the resource commitments of a project
- will be used to calculate how many resources are commited but not used

---

- currently I am just returning 0 capacity if a flavor can not be mapped, do we need to take care of this?
- what about disabled projects?


PhilippMatthes (commented):
Can't we use the nova limits api to fetch the data you're manually extracting here?

PhilippMatthes:
Clarified this offline

PhilippMatthes (approved):

PR #299, opened 2025-09-18
Add commitment change api for limes nova liquid

PhilippMatthes:
Limes can now delegate commitment acceptance to an external service in its liquids. This implements a new endpoint that takes a commitment change request, checks it against the initial placement pipeline (if there is a valid host) and rejects it if not.

WIP


internal/scheduler/nova/api/http/api.go
```diff
@@ -168,3 +181,209 @@ func (httpAPI *httpAPI) NovaExternalScheduler(w http.ResponseWriter, r *http.Req
...
+
+			nNewVMs := int(totalAfter) - int(totalBefore)
+			if nNewVMs <= 0 {
+				slog.Debug("won't reject commitment conversion/shrinkage", "nNewVMs", nNewVMs)
+				continue
+			}
+			flavorName := strings.TrimPrefix(resourceName, "instances_")
+			flavor, ok := flavorsByName[flavorName]
+			if !ok {
+				slog.Warn("won't reject unknown flavor in commitment change", "flavorName", flavorName)
```
SoWieMarkus, line 316: Why? I mean we shouldn't just accept it?

PhilippMatthes: I'll make a hard constraint out of it and add a code comment.

PR #306, opened 2025-09-24
Refactor syncing of flavors, servers, and deleted servers

SoWieMarkus:
I noticed some data inconsistencies between two deployments that should have exactly the same data—they report a different number of `ACTIVE` servers.

While investigating, @PhilippMatthes and I identified the changes-since parameter in the OpenStack servers syncer as the most likely reason for this issue. We discussed whether we actually need this parameter, since it is only used for the VM life span KPI. We decided to remove it for now. While doing this, we made the same change for all the other syncers.

Let’s use this PR to discuss the topic. @auhlig 

Example query:
```sql
-- Get all servers that are not deleted (can be ERROR, SHUTOFF, ...)
SELECT COUNT(*) AS total_not_deleted FROM openstack_servers WHERE status != 'DELETED';
-- Get all active servers
SELECT COUNT(*) AS total_active FROM openstack_servers WHERE status = 'ACTIVE';
```


PhilippMatthes:
Markus and I discussed it would be good to have 2 tables, 1 for deleted and 1 for active servers. The deleted table should fetch all servers from the origin of time.

PhilippMatthes (approved):
Nice!

PR #320, opened 2025-10-07
Box out scheduler

PhilippMatthes:
We want to box out the services currently contained in internal/ into separate code paths + binaries + charts. This PR starts with the scheduler.

Change summary
- Move internal/scheduler code to  scheduler/
- Move scheduler-specific configuration from internal/conf to scheduler/
- Move e2e tests for scheduler from commands/checks to scheduler/
- Move workload spawner from commands/spawner up to spawner/
- Add new scheduler dist chart, use it in the bundles
- Remove scheduler mode from cortex-core library
- Share global configuration with global helm values
- Adjust the tilt setup to the changed structure
- Move methods of the external scheduler requests needed by the pipeline into the pipeline api code

Other changes:
- Rename NewConfig to GetConfigOrDie

Note after this change we will need to update our production setups to the changed helm values structure, as well as the ci pipeline to use the changed e2e tests command. We will also need to update our cortex.secrets.yaml changing `cortex-core:` to `global:`.


SoWieMarkus (approved):
Nice! Looking forward to the other charts! I didn't notice anything that isn't working anymore

PR #354, opened 2025-10-27
Visualizer use kubectl proxy

SoWieMarkus:
Visualizer now connects to the local kubernetes api server via `kubectl proxy`. See [here](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_proxy/) and [here](https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/)

### DIFF
--- a/visualizer/nova.html
+++ b/visualizer/nova.html
@@ -276,7 +276,6 @@
 
       listSchedulingDecisions(vmId)
         .then(decision => {
-          console.log(decision)
           if (!decision) {
             alert(`No scheduling decision found for vm_id ${vmId}`)
             return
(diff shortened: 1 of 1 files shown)


PhilippMatthes (commented):
Super cool!

Would be nice to use a bearer token and directly call the apiserver without kubectl-proxy. I think this can be achieved as follows:

(In Tilt)
- Create serviceaccount, let's say `cortex-visualizer`
- Create clusterrolebinding `cortex-visualizer-crb` for sa binding role `cluster-admin`
- Create bearer token for the serviceaccount, maybe 4 weeks (irrelevant for local debugging, but just to be safe and not configure its lifetime forever)
- Using kubectl get the apiserver url
- Proxy apiserver over nginx (but use the url provided before by kubectl)
- Put the bearer token into the json for the visualizer build

(In visualizer)
- Call nginx proxy with bearer token as header to authenticate

Result should be: we can access all resources we want in the visualizer

PR #355, opened 2025-10-29
Reintroduce explanation controller (why/why not api)

PhilippMatthes:
Change summary:
- Reconciles history of decision objects
- Reconciles explanation of decision objects


scheduling/internal/e2e/nova/checks.go
```diff
@@ -132,7 +132,7 @@ func prepare(ctx context.Context, config conf.Config) datacenter {
 	slog.Info("found hypervisors", "count", len(hypervisors))
 
 	slog.Info("listing servers")
-	slo := servers.ListOpts{AllTenants: true, Limit: 100}
+	slo := servers.ListOpts{AllTenants: true, Limit: 10}
```
PhilippMatthes: This is so that we have some requests for the same server id

PR #385, opened 2025-11-20
Implement alerts for cortex services

SoWieMarkus:
## Description

With the new repository structure and configuration of the bundles, the Cortex microservices are now much more flexible. Since we configure the ServiceMonitor for each bundle individually, it makes sense to configure alerts per bundle as well.

1. We've unified several microservices (e.g., `extractor` + `kpi` + `syncer` → `knowledge`), significantly reducing the number of alerts needed
2. Different bundles have different components (e.g., `cortex-ironcore` doesn't have a knowledge module at all)
3. While unifying alerts across bundles is technically possible, it would create configuration complexity. Managing alerts per bundle is more straightforward and easier to maintain

## Changes

- Removed the centralized `cortex-alerts` chart
- Added bundle-specific alert configurations
- Scoped alerts to their respective bundles (nova, cinder, manila, ironcore)
- Added github workflow to check the alerts

### DIFF
--- a/.github/workflows/check-alerts.yaml
+++ b/.github/workflows/check-alerts.yaml
@@ -11,15 +11,18 @@ jobs:
     steps:
       - name: Checkout PR
         uses: actions/checkout@v5
+
       - name: Get changed rule and alert files
         id: changed
         uses: tj-actions/changed-files@v46
-      - name: Install Helm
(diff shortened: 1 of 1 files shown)


PhilippMatthes (approved):
Thanks so much!

PR #395, opened 2025-11-26
Disable decision creation in pipeline configurations

SoWieMarkus:


helm/bundles/cortex-nova/values.yaml
```diff
@@ -59,6 +59,9 @@ openstack:
     enabled: false
     <<: *sharedSSOCert
 
+pipelines:
+  createDecisions: false
```
PhilippMatthes:
```suggestion
  # Use this flag to disable the creation of decisions across all pipelines.
  createDecisions: false
```

PhilippMatthes (approved):
Thanks!

Issue #392, opened 2025-11-24
Drop-in Kubernetes Scheduler


auhlig:
### Description

For Kubernetes-based GPU workloads, Cortex should act as a drop-in scheduler for annotated pods (`.spec.schedulerName="cortex"`).
A dedicated Kubernetes integration should watch such pods, translate their scheduling requirements (incl. GPU-related constraints) into Cortex requests, and bind them to suitable nodes based on Cortex’ decisions.
The logic should be abstracted via a plugin in Cortex.

### Objectives

- Implement a controller that watches pods with `.spec.schedulerName="cortex"`
- Translate pod resource requests (incl. GPU resources, affinities, tolerations) into Cortex scheduling requests
- Bind pods to nodes based on Cortex decisions
- Handle failure modes cleanly (e.g., Cortex unavailable → pods remain Pending)
- Provide minimal metrics and logs
- Documentation

### Acceptance Criteria

- [ ] Pod controller implemented
- [ ] Pods annotated with `.spec.schedulerName="cortex"` are scheduled by Cortex only
- [ ] Basic e2e tests verify behavior

### Dependencies

N/A

### Additional Notes

N/A

PR #406, opened 2025-12-01
Implement Cortex support for Pods

kruegercharles:
Ref-Issue: #392


SoWieMarkus:
Hey Charly 👋

I noticed your test workflow is failing when trying to post the coverage comment. This happens because PRs from external contributors don't have write permissions to post comments (which is a GitHub security feature).

I've fixed this by adding `continue-on-error: true` to the comment steps, so the workflow will now succeed even if it can't post the comment. The tests will still run normally and you'll see the results in the workflow logs.

Just rebase on `main` to pick up the updated workflow.

internal/scheduling/decisions/pods/pipeline_controller.go
```diff
@@ -0,0 +1,308 @@
...
+			c.handlePod(),
+			// Only schedule pods that have a custom scheduler set.
+			predicate.NewPredicateFuncs(func(obj client.Object) bool {
+				pod := obj.(*corev1.Pod)
+				if pod.Spec.NodeName != "" {
+					// Skip pods that already have a node assigned.
+					return false
+				}
+				// FIXME: is this correct?
+				return pod.Spec.SchedulerName == c.Conf.Operator
```
SoWieMarkus: Did you check it? For me the approach looks good :D If so remove the comment

kruegercharles: Oh, I forget to remove the comment. I removed it now.

I also changed the scheduler name from `cortex-pods` to `cortex` to conform with Arno's proposal.

I've just tested it and the scheduler name gets set correctly for the test-pod.

config/crd/cortex.dev_nodes.yaml
```diff
@@ -0,0 +1,551 @@
...
+              volumesInUse:
+                description: List of attachable volumes in use (mounted) by the node.
+                items:
+                  type: string
+                type: array
+                x-kubernetes-list-type: atomic
+            type: object
+        type: object
+    served: true
+    storage: true
```
PhilippMatthes: What's this CRD?

helm/bundles/cortex-pods/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+name: cortex-pods
+description: A Helm chart deploying Cortex for Pods.
+type: application
+version: 0.0.1
+appVersion: 0.1.0
+dependencies:
+  # from: file://../../../dist/chart
+  - name: cortex
+    repository: oci://ghcr.io/cobaltcore-dev/cortex/charts
+    version: 0.0.2
```
PhilippMatthes: https://github.com/cobaltcore-dev/cortex/blob/f80107f41aa15ab636b6db86944a161f51d320bc/dist/chart/Chart.yaml#L5

kruegercharles: Done 👍

config/crd/cortex.dev_nodes.yaml
kruegercharles: Those crds were from an earlier version where I (redundantly) added custom crds for nodes and pods. 
Now we just use the builtin nodes and pods from Kubernetes.

I deleted them.

PR #456, opened 2026-01-12
Affinity taint toleration filters

Varsius:
This PR refactors the directory structure for filters/weighers of the pod pipeline to align with the nova and manila pipeline. Secondly, filters are added that implement the default Kubernetes behavior for hard taints, tolerations and affinities. Soft requirements regarding these behaviors, which would need to be implemented as weighers, are not included in this PR and are subject for future work.

### DIFF
--- a/internal/scheduling/decisions/pods/plugins/filters/filter_node_affinity.go
+++ b/internal/scheduling/decisions/pods/plugins/filters/filter_node_affinity.go
@@ -29,7 +29,7 @@ func (NodeAffinityFilter) Run(traceLog *slog.Logger, request pods.PodPipelineReq
 
 	for _, node := range request.Nodes {
 		if matchesNodeAffinity(node, request.Pod) {
-			activations[node.Name] = 1.0
+			activations[node.Name] = 0.0
 		}
 	}
 
--- a/internal/scheduling/decisions/pods/plugins/filters/filter_taint.go
+++ b/internal/scheduling/decisions/pods/plugins/filters/filter_taint.go
@@ -28,7 +28,7 @@ func (TaintFilter) Run(traceLog *slog.Logger, request pods.PodPipelineRequest) (
 
 	for _, node := range request.Nodes {
 		if canScheduleOnNode(node, request.Pod) {
-			activations[node.Name] = 1.0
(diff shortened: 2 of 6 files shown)


internal/scheduling/decisions/pods/plugins/filters/filter_node_affinity.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+		return true
+	case corev1.NodeSelectorOpExists:
+		return exists
+	case corev1.NodeSelectorOpDoesNotExist:
+		return !exists
+	case corev1.NodeSelectorOpGt:
+		return exists && nodeValue > req.Values[0]
+	case corev1.NodeSelectorOpLt:
+		return exists && nodeValue < req.Values[0]
```
PhilippMatthes: Is this safe?

internal/scheduling/decisions/pods/plugins/filters/filter_taint.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (TaintFilter) Run(traceLog *slog.Logger, request pods.PodPipelineRequest) (*lib.StepResult, error) {
+	activations := make(map[string]float64)
+	stats := make(map[string]lib.StepStatistics)
+
+	for _, node := range request.Nodes {
+		if canScheduleOnNode(node, request.Pod) {
+			activations[node.Name] = 1.0
+		}
```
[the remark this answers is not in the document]
PhilippMatthes: Technically, we're not too precise with the wording here. The `activations` map is actually a `weights` map which is going to be mapped to an activation by the activation function. We should rename this.

internal/scheduling/decisions/pods/plugins/filters/filter_node_affinity.go
Varsius: Good catch! I also completely missed the parsing to integers here. I now implemented these cases according to the [Kubernetes Affinity Operator Documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#operators) and added tests for these.

PhilippMatthes (approved):

PR #461, opened 2026-01-20
Add binpack weigher for pods pipeline

Varsius:


internal/scheduling/decisions/pods/plugins/weighers/binpack.go
```diff
@@ -0,0 +1,79 @@
...
+	podResources := shared.GetPodResourceRequests(request.Pod)
+
+	for _, node := range request.Nodes {
+		result.Activations[node.Name] = s.calculateBinpackScore(node, podResources, s.Options.ResourceWeights)
+	}
+
+	return result, nil
+}
+
+func (s *BinpackingStep) calculateBinpackScore(node corev1.Node, podResources corev1.ResourceList, weights map[corev1.ResourceName]float64) float64 {
```
PhilippMatthes: Did I understand it correctly?
```suggestion
// calculateBinpackScore outputs a higher score if the 
// node allocation is closer to 100% after the pod has
// been placed.
func (s *BinpackingStep) calculateBinpackScore(node corev1.Node, podResources corev1.ResourceList, weights map[corev1.ResourceName]float64) float64 {
```

PR #495, opened 2026-02-05
Use gvk in multicluster client to map resources

PhilippMatthes:
Before, it was necessary for resources to declare a `URI()` method to be registered in the multicluster client. Now, this is handled over the group version kind (gvk), allowing us to also handle resources outside of our own packages with the multicluster client, such as the hypervisor crd. This makes use of the fact that each type handled by the controller-runtime client conforms to the runtime.Object interface, providing the gvk over `GetObjectKind`.

### DIFF
--- a/pkg/multicluster/builder.go
+++ b/pkg/multicluster/builder.go
@@ -35,6 +35,10 @@ type MulticlusterBuilder struct {
 //
 // If the object implements Resource, we pick the right cluster based on the
 // resource URI. If your builder needs this method, pass it to the builder
+// Watch resources, potentially in a remote cluster.
+//
+// Determines the appropriate cluster by looking up the object's GroupVersionKind (GVK)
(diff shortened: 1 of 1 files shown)


pkg/multicluster/client.go
```diff
@@ -79,104 +98,94 @@ func (c *Client) ClusterForResource(uri string) cluster.Cluster {
...
+	if err != nil {
+		return err
 	}
-	cl := c.ClusterForResource(resource.URI())
-	return cl.GetClient().Create(ctx, obj, opts...)
+	return c.ClientForResource(gvk).Create(ctx, obj, opts...)
 }
 
-// Pick the right cluster based on the resource URI and perform a Create operation.
+// Pick the right cluster based on the resource type and perform a Create operation.
```
SoWieMarkus:
```suggestion
// Pick the right cluster based on the resource type and perform a Delete operation.
```

Issue #497, opened 2026-02-05
[Feature Idea] Generic Prometheus Metric Plugin for Dynamic Weigher/Filter Logic


henrichter:
_While working on energy awareness the following idea came up._
### Description

**Summary**

Add support for a "Generic Metric" type that allows users to define complex Prometheus queries directly in the Datasource CR and use the results for weighing or filtering hosts during scheduling decisions. This would enable building weigher/filter logic purely in Prometheus queries without requiring custom Go code for each new metric.

---

### Motivation

Currently, adding a new Prometheus-based weigher or filter requires:
1. Creating a new typed metric struct in `internal/knowledge/datasources/plugins/prometheus/types.go`
2. Implementing a new extractor in `internal/knowledge/extractor/plugins/`
3. Implementing a new weigher/filter in `internal/scheduling/<domain>/plugins/weighers/`
4. Registering the new plugin in the index

This is a significant amount of boilerplate for what is sometimes a simple "query Prometheus -> map to hosts -> apply weight/filter" pattern. A generic implementation would allow operators to:
- Rapidly prototype new scheduling heuristics
- Use complex PromQL aggregations without code changes
- Experiment with different metrics without redeploying cortex

---

#### Example 1: Weigher - Prefer Hosts with Low CPU Usage

**Goal:** Prefer hosts where the CPU is mostly idle (high idle ratio).

**Prometheus Query:**
```promql
1 - (
  avg by (instance) (rate(node_cpu_seconds_total{mode!="idle"}[5m]))
  /
  avg by (instance) (rate(node_cpu_seconds_total[5m]))
)
```

This returns a value between 0 and 1, where 1 = fully idle, 0 = fully utilized.

**Desired Behavior:**
- Higher values (more idle) -> higher weight
- This would be configured as a weigher in the pipeline

**Example Datasource CR:**
```yaml
apiVersion: cortex.cloud/v1alpha1
kind: Datasource
metadata:
  name: node-cpu-idle-ratio
spec:
  schedulingDomain: nova
  type: prometheus
  prometheus:
    query: |
      1 - (
        avg by (instance) (rate(node_cpu_seconds_total{mode!="idle"}[5m]))
        /
        avg by (instance) (rate(node_cpu_seconds_total[5m]))
      )
    alias: node_cpu_idle_ratio
    type: generic  # new
```

#### Example 2: Filter - Exclude Hosts Under IRQ Pressure

[...]

PR #532, opened 2026-02-19
feat: add generic metric, extractor, weigher/filter

henrichter:
Implements the idea of #497

I'll add some examples soon


PhilippMatthes:
Thank you! 

There are surely some things we can improve with the knowledge database. Your contribution fits exactly where we see room for improvement. Removing complexity, indirections, and ultimately the postgres dependency from the cortex knowledge database.

I would think even more broadly: Ideally, we don't even have to store this data inside an additional database, and just use whatever comes from prometheus, aggregating it; for example; inside the knowledge crd. In that sense, our metrics caching layer is not needed necessarily, and the knowledge is by definition a condensed async picture of the datacenter state.

Furthermore, PromQL should provide the necessary means to pull something like a "generic metric calculation result" like you implemented here; but then I would directly continue to process it, instead of putting it into another persistence layer. Maybe we can also design the knowledge crd and controller logic generic enough to support ingesting data from various sources (up for discussion). So: there are lots of ideas here up for grabs!

Finally, sorry for leaving your contribution untouched for the last couple months. The core idea is good and we keep it as a reference inside the git tree to revisit as part of abovementioned efforts.

PR #610, opened 2026-03-23
Liquid endpoint report capacity

juliusclausnitzer:


internal/scheduling/reservations/commitments/capacity.go
```diff
@@ -96,30 +118,156 @@ func (c *CapacityCalculator) copyAZCapacity(
...
+	}
+
+	azs := make([]string, 0, len(azSet))
+	for az := range azSet {
+		azs = append(azs, az)
+	}
+	sort.Strings(azs)
+
+	return azs, nil
+}
```
SoWieMarkus: Some comments on that:

1. Why do we first create a host to az map that we then reduce to an AZ set in another function. Wouldn't it make much more sense to have that in one function? Or do we need that map somewhere else.

2. Not very obvious by the name (will change that), but the Host Details Knowledge was basically my approach to have an VMware hypervisor CRD. All the hosts inside this knowledge only refer to vmware hosts! Since this feature is intended to be used for kvm I would suggest that we use the Hypervisor CRD for this. The availability zone there is available via a label (`topology.kubernetes.io/zone`).

internal/scheduling/reservations/commitments/capacity.go
```diff
@@ -6,28 +6,39 @@ package commitments
...
+		totalPipeline:   config.ReportCapacityTotalPipeline,
+	}
 }
 
 // CalculateCapacity computes per-AZ capacity for all flavor groups.
 // For each flavor group, three resources are reported: _ram, _cores, _instances.
 // All flavor groups are included, not just those with fixed RAM/core ratio.
-// The request provides the list of all AZs from Limes that must be included in the report.
-func (c *CapacityCalculator) CalculateCapacity(ctx context.Context, req liquid.ServiceCapacityRequest) (liquid.ServiceCapacityReport, error) {
+// AZs are derived from HostDetails Knowledge CRDs.
```
SoWieMarkus: See comment about the host details knowledge

internal/scheduling/reservations/commitments/capacity.go
```diff
@@ -96,30 +118,156 @@ func (c *CapacityCalculator) copyAZCapacity(
...
+		MemoryMB:         smallestFlavor.MemoryMB,
+		VCPUs:            smallestFlavor.VCPUs,
+		FlavorExtraSpecs: map[string]string{"hw_version": groupName},
+		AvailabilityZone: az,
+		Pipeline:         c.totalPipeline,
+	})
+	if err != nil {
+		return 0, fmt.Errorf("scheduler call failed: %w", err)
+	}
+
```
SoWieMarkus, line 195: Lets have another iteration about this in the tech sync. It feels very wrong to call our external scheduler api for something like that. We should be able to directly use the pipeline here

SoWieMarkus:
Can you add a description of your changes to the PR?

internal/scheduling/reservations/commitments/capacity.go
juliusclausnitzer: Does this look better to you now?

internal/scheduling/reservations/commitments/capacity.go
PhilippMatthes: Decision controller + decision crd? ;)

PhilippMatthes:
```yaml
apiVersion: cortex.cloud/v1alpha1
kind: Decision
metadata:
  name: reservation-candidates-<uuid>
spec:
  constraints:
    affinityRules:
      # Reservation on same host as reference vm?
      # (for failover)
      - avoid: {vm: {uuid: "<vmuuid>"}}
    flavor: 
      name: "hana-small"
      extraSpecs: {} # ...
    resources:
      memory: "1Gi"
      cpu: "128"
# After reconciliation by decision controller
status:
  candidates: ["host-1", "host-2", "host-3"] 
  conditions:
    - type: "Ready"
      status: "True"
```

PR #640, opened 2026-03-26
[after GA] Commitments syncer and change-api share mutex lock

mblos:


PhilippMatthes (changes requested):

mblos:
close for now. components will be refactored otherwise

PR #688, opened 2026-04-08
Improve logging and  add metrics for better ops experiance

umswmayj:
- Production-safe zap defaults (JSON, no multiline stack traces) + LOG_LEVEL env var for slog
- Single logLevel Helm value controls both zap and slog log level
- 5 reconcile logs downgraded to debug, 2 errors downgraded to warn
- Fix maps.Keys() emitting !ERROR:json: unsupported type: iter.Seq[string]
- New cortex_log_messages_total{level, file} counter for both slog and zap loggers


helm/bundles/cortex-nova/alerts/nova.alerts.yaml
```diff
@@ -157,7 +157,7 @@ groups:
         time the service will have a less recent view of the datacenter.
 
   - alert: CortexNovaSyncObjectsDroppedToZero
-    expr: cortex_sync_objects{service="cortex-nova-metrics", datasource!="openstack_migrations"} == 0
+    expr: cortex_sync_objects{service="cortex-nova-metrics", datasource!~"openstack_migrations|prometheus_kvm_libvirt_domain_steal_pct"} == 0
```
PhilippMatthes: Silence the alert instead?

umswmayj: Maybe also an option. The issue in general is that not all datasource (e.g. prometheus_kvm_libvirt_domain_steal_pct) does not always have `FEATURES > 0`

umswmayj:
Will split PR => close

PR #691, opened 2026-04-09
Scaffold placement api shim

PhilippMatthes:
This adds the initial scaffold for the placement api shim. 

The shim's main entrypoint is placed under `cmd/shim/main.go` moving the original entrypoint to `cmd/manager/main.go`. This allows the shim and manager to be built into two separate images with different release cycles than the main controller-manager binary. Which binary is built and packaged into the docker image is controlled via a new `GOMAIN` build arg.

To onboard other shims in the future, such as a shim for our limes API or a shim for the nova external scheduler api, `internal/shim` and `cmd/shim` is the place where the code should live. A `helm/library/cortex-shim` provides the necessary scaffold to deploy any shim. We follow the same pattern as with the cortex manager chart, bundling this under `helm/bundles/cortex-placement-shim` with the necessary customization: placement-shim specific alerts, config, and rbac.

Completing this scaffold, the image and charts are added to our gh workflow and the Tiltfile. Also, a basic kubebuilder manager scaffold has been added to `cmd/shim/main.go`.


.github/workflows/push-charts.yaml
```diff
@@ -27,25 +27,6 @@ jobs:
...
-        env:
-          ALL_CHANGED_FILES: ${{ steps.changed-chart-yaml-files-core.outputs.all_changed_files }}
-        run: |
-          for CHART_FILE in ${ALL_CHANGED_FILES}; do
-            CHART_DIR=$(dirname $CHART_FILE)
-            helm package $CHART_DIR --dependency-update --destination $CHART_DIR
-            CHART_PACKAGE=$(ls $CHART_DIR/*.tgz)
-            helm push $CHART_PACKAGE oci://${{ env.REGISTRY }}/${{ github.repository }}/charts/
-          done
-
```
PhilippMatthes, line 48: This workflow step is redundant with the one below. It can be removed.

cmd/shim/main.go
```diff
@@ -0,0 +1,252 @@
...
+		// Kept for consistency with kubebuilder scaffold, but the shim should
+		// always run with leader election disabled.
+		LeaderElection: enableLeaderElection,
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	// TODO: Initialize multicluster client here.
```
SoWieMarkus, line 198: To make this easier in the future: https://github.com/cobaltcore-dev/cortex/pull/702

cmd/shim/main.go
```diff
@@ -0,0 +1,252 @@
...
+		os.Exit(1)
+	}
+
+	errchan := make(chan error)
+	go func() {
+		errchan <- func() error {
+			setupLog.Info("starting api server", "address", ":8080")
+			return httpext.ListenAndServeContext(ctx, ":8080", mux)
+		}()
+	}()
```
SoWieMarkus: Nit: We could run the Cortex API servers as a controller-runtime runnable so it shares the managers lifecycle / behavior on error. I guess that would a little bit "cleaner". This would also apply for the external scheduler api for nova/cinder/manila etc.

cmd/shim/main.go
PhilippMatthes: See https://github.com/cobaltcore-dev/cortex/pull/702#pullrequestreview-4097058898
> Please note these changes I made as part of my multicluster-integration for the placement shim. It's probably better than making an assumptive `NewClient` method: https://github.com/cobaltcore-dev/cortex/blob/5bedfb5ba06785adaa6bd312239645100732004c/pkg/multicluster/routers.go#L15-L22
> Usage: https://github.com/cobaltcore-dev/cortex/blob/5bedfb5ba06785adaa6bd312239645100732004c/cmd/shim/main.go#L220-L225

SoWieMarkus (approved):

cmd/shim/main.go
PhilippMatthes: Good point. I think it makes sense to start the mux as part of the manager lifecycle as soon as the cache is available. This would make a healthcheck guiding traffic onto the mux only after cache initialization obsolete. I will code this in, and afterward adjust my follow-up PR accordingly.

PhilippMatthes: Resolved in https://github.com/cobaltcore-dev/cortex/pull/691/commits/cc144b1a406da0ae7e83b802e35070d2711c8eb8

PR #692, opened 2026-04-09
Add passthrough handlers and mcl setup to placement shim

PhilippMatthes:
In #691 we scaffolded the helm chart and entrypoint for the placement api shim. Now, we add a `type Shim struct` which carries all http handlers needed to serve the full placement api. This shim needs to talk in two directions: 1. to kubernetes using a multicluster controller-runtime client to access the hypervisor crd and 2. to real placement via a http client. 

To support (1), a multicluster controller-runtime client is now initialized wired with `type Shim struct` such that the hypervisor crd informer-cache gets initialized. The easiest way to integrate this with our multicluster setup is to implement `type Shim struct` as a controller. To implement (2), a http client is initialized with proper transport parameters to avoid foreseeable issues like hanging handlers, using timeouts and increasing the number of parallel connections. The connection is initialized using `conf.placementURL` on controller startup and then tested on placement's `/` endpoint.


PhilippMatthes:
^ Ignored the comments already discussed in #691 or related to the kubebuilder scaffold.

PR #698, opened 2026-04-10
Number of minor ops cleanups

umswmayj:


.github/renovate.json
```diff
@@ -40,6 +40,19 @@
...
+      "matchStrings": ["FROM (?<depName>[^:\\n]+):(?<currentValue>[^@\\n]+)@sha256:(?<currentDigest>[a-f0-9]+)"],
+      "datasourceTemplate": "docker"
+    },
+    {
+      "fileMatch": ["^postgres/Dockerfile$"],
+      "matchStrings": ["ENV PG_VERSION (?<currentValue>[\\d]+\\.[\\d]+)-[^\\n]+"],
+      "depNameTemplate": "postgres",
+      "datasourceTemplate": "docker",
+      "versioningTemplate": "semver-coerced",
+      "autoReplaceStringTemplate": "ENV PG_VERSION {{{newValue}}}-1.pgdg13+1"
```
SoWieMarkus, line 55: Love this! And since we have a workflow that automatically updates the appVersion of the postgres chart this would also result in a `helm-lint` error when we are releasing a new version and forgot to update the postgres chart version.

postgres/Dockerfile
```diff
@@ -1,5 +1,5 @@
-# Last updated: 17 Mar 2026
-FROM debian:trixie-slim@sha256:26f98ccd92fd0a44d6928ce8ff8f4921b4d2f535bfa07555ee5d18f61429cf0c
+# Last updated: 9 Apr 2026
```
SoWieMarkus: We used this comment before to trigger a new image build if there was an upstream fix and nothing to do for us. Since we have the renovate PR now (which doesn't update this comment) with the fixed sha we can get rid of this comment.

SoWieMarkus (approved):

PR #699, opened 2026-04-10
Feat: Add logging config option to helm

umswmayj:


SoWieMarkus (approved):
See coderabbit comments

PR #700, opened 2026-04-10
feat: add metrics for number of warnings / errors

umswmayj:


pkg/monitoring/log_metrics.go
```diff
@@ -0,0 +1,124 @@
...
+// this map grows to a fixed size and all subsequent lookups are lock-free reads.
+var pcFileCache sync.Map // uintptr -> string
+
+// LogMessagesTotal counts warn and error log messages emitted by both the slog
+// and zap loggers. Labels: "level" (warn|error), "file" (relative source path).
+var LogMessagesTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "cortex",
+	Name:      "log_messages_total",
+	Help:      "Total number of log messages emitted at warn or error level.",
+}, []string{"level", "file"})
```
SoWieMarkus: In other parts of cortex we always designed a `Monitor` which owns the the metric. We _could_ add this pattern here as well (add the metric to the MetricsSlogHandler, maybe rename it to SlogHandlerMonitor) which implements the Monitor interface see [e.g. here](https://github.com/cobaltcore-dev/cortex/blob/main/internal/knowledge/datasources/monitor.go) to be consistent. But not strong opinion against your approach

SoWieMarkus (approved):

PR #716, opened 2026-04-16
fix: remove unused db connection

mblos:


SoWieMarkus (changes requested):
Can you remove the db secrets reference in the values as well?

SoWieMarkus:
Sorry misclicked and accidentally requested the review of copilot

cmd/manager/main.go
```diff
@@ -360,7 +360,11 @@ func main() {
 
 	// Initialize commitments API for LIQUID interface (Postgres-backed usage reporting).
 	commitmentsConfig := conf.GetConfigOrDie[commitments.Config]()
-	commitmentsAPI := commitments.NewAPIWithConfig(multiclusterClient, commitmentsConfig, nil)
+	var commitmentsUsageDB commitments.UsageDBClient
+	if commitmentsConfig.DatasourceName != "" {
+		commitmentsUsageDB = commitments.NewDBUsageClient(multiclusterClient, commitmentsConfig.DatasourceName)
+	}
```
SoWieMarkus, line 366: What happens if the datasource name is not set? Could this lead to a crash loop of the service if commitmentsUsageDB is nil?

SoWieMarkus (approved):

cmd/manager/main.go
mblos: should™️ be safe.. and API returns then http 500

PR #728, opened 2026-04-22
feat: Add FlavorGroupCapacity CRD and Capacity Controller

juliusclausnitzer:
Introduces a `FlavorGroupCapacity` CRD and a background capacity controller that pre-computes per-flavor VM slot capacity for each (flavor group × AZ) pair on a fixed interval. This allows future capacity API endpoints to serve data from cache instead of probing the scheduler on every request.

## CRD (`api/v1alpha1/FlavorGroupCapacity`)
One cluster-scoped resource per (flavor group × AZ). Status holds per-flavor fields: `totalCapacityVmSlots` / `totalCapacityHosts` (empty-datacenter scenario via `kvm-report-capacity` pipeline) and `placeableVms` / `placeableHosts` (current-state via `kvm-general-purpose-load-balancing`), plus `totalInstances` and `committedCapacity` aggregated from `CommittedResource` CRDs.

## Controller (`internal/scheduling/reservations/capacity/`)
Runs as a `manager.Runnable` on a configurable interval (default 5 min). For each (group × AZ) pair it issues two scheduler probes — one per pipeline. VM slots are computed as `floor(host.effectiveCapacity.memory / flavor.memoryBytes)`. Stale per-flavor values are preserved on partial probe failure; the `Ready` condition reflects whether all probes succeeded.

## Wiring
`capacity-controller` added to `enabledControllers` in the Helm chart values. `FlavorGroupCapacity` and `FlavorGroupCapacityList` registered as home-cluster GVKs in the multicluster client config.


juliusclausnitzer:
@coderabbitai review

mblos:
small comments:
- limitHostsToRequest restricts in the end to the candidate hosts..so we should pass over candidates when doing the scheduling call
- flavor group names for CRD naming.. what when they differ only by eg . vs - produce the same name. Suggested approach: "petname" style. append a short hash of the raw (unsanitized) inputs to the sanitized prefix, e.g. hana-v2-qa-de-1a-a3f9c2. This keeps names human-readable while making them collision-free .. and some cap to be within the char limit. Adding some test explicitly for such similar named crds would be great
 - for the CRD format.. I think we should have in the status for each of the flavors of the flavor group the stats in there. that would be helpful for the dashboard. wdyt?

CRD format
```
status:
  flavors:
    - flavorName: hana-v2-small
      placeableHosts: 38        # current state, hosts that fit this flavor
      placeableVms: 95          # current state, VM slots remaining
      totalCapacityHosts: 40    # empty state, matching hosts
      totalCapacityVmSlots: 120 # empty state, max VM slots
    - flavorName: hana-v2-large
      ...
  committedCapacity: 15   # sum of AcceptedAmount across active CommittedResource CRDs
  totalInstances: 310     # running VMs in this AZ (from HV CRDs, not flavor-filtered)
  lastReconcileAt: ...
  conditions: [Ready]
  ```

PR #730, opened 2026-04-23
feat: adding CommittedResource CRD

mblos:
Introduces CommittedResource CRD for commitments of various resource types (memory, cpu, ...). This CRD acts as the API how to configure commitments in Cortex.

Related: https://github.wdf.sap.corp/cc/secrets/pull/20201


mblos:
Related: https://github.wdf.sap.corp/cc/secrets/pull/20201

api/v1alpha1/committed_resource_types.go
```diff
@@ -0,0 +1,179 @@
...
+// CommittedResourceSpec defines the desired state of CommittedResource,
+type CommittedResourceSpec struct {
+	// UUID of the commitment this resource corresponds to.
+	// +kubebuilder:validation:Required
+	CommitmentUUID string `json:"commitmentUUID"`
+
+	// SchedulingDomain specifies the scheduling domain for this committed resource (e.g., "nova", "ironcore").
+	// +kubebuilder:validation:Enum=nova;cinder;manila;machines;pods
+	// +kubebuilder:validation:Required
+	SchedulingDomain SchedulingDomain `json:"schedulingDomain"`
```
juliusclausnitzer: ironcore is explicitly mentioned as example but not a possible input value given validation

api/v1alpha1/committed_resource_types.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// ResourceType identifies the kind of resource committed: memory drives Reservation slots; cores uses an arithmetic check only.
+	// +kubebuilder:validation:Enum=memory;cores
+	// +kubebuilder:validation:Required
+	ResourceType CommittedResourceType `json:"resourceType"`
+
+	// Amount is the total committed quantity.
+	// Examples: "1280Gi" for RAM, "40" for CPU cores.
+	// +kubebuilder:validation:Required
+	Amount resource.Quantity `json:"amount"`
```
juliusclausnitzer, line 62: do we not work in multiples of smallest flavor here?

juliusclausnitzer, line 62: Also, does this mean that we would create multiple committed resources for one reservation so we can keep track of both cpu cores and RAM? 

I would have expected something more like this, but maybe I am off-track here:

```// Resources defines the set of committed resources (e.g. cpu, memory).
	// Example:
	// cpu: "40"
	// memory: "1280Gi"
	// +kubebuilder:validation:Required
	Resources corev1.ResourceList `json:"resources"`
```

api/v1alpha1/committed_resource_types.go
mblos: Good catch, I removed the validation.

api/v1alpha1/committed_resource_types.go
mblos: The CRD should be self contained, so not depending implicitly on e.g. flavor group definition, also considering that we were told the groups are likely to change afterwards. So the Liquid API will do the translation and writes the raw amount here

mblos: With Limes we have always one resource per commitment. So we have e.g. two commitment for the same project, when the user wants to specify memory and cpu for a GP group. Still our controller can read both related CR CRDs to do more advanced flavor selection.. considering the CPU cores the user has in mind while choosing the right flavor sizes for the reservations of the memory cr crd.
I also thought about using a ResourceList with always using a single resource in there.. but it feels like doing optimization for a use case we dont have atm.

PR #766, opened 2026-04-28
Support traits and aggregates endpoints in placement shim

PhilippMatthes:
Implements GET/PUT/DELETE /resource_providers/{uuid}/traits and GET/PUT /resource_providers/{uuid}/aggregates with three feature modes (passthrough, hybrid, crd-only). Hybrid mode serves from the Hypervisor CRD for KVM providers and forwards to upstream placement for non-KVM. CRD-only mode returns 404 for non-KVM providers. Uses metadata.generation for optimistic concurrency (409 on mismatch). Includes unit tests and e2e tests for all modes.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -83,7 +83,7 @@ local('kubectl wait --namespace cert-manager --for=condition=available deploymen
 
 ########### Dependency CRDs
 # Make sure the local cluster is running if you are running into startup issues here.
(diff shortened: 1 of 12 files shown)


PhilippMatthes:
See https://github.com/cobaltcore-dev/openstack-hypervisor-operator/pull/292

PhilippMatthes:
e2e tests are still broken because we need to merge the hypervisor crd first and then update it in the cortex repo.

PR #767, opened 2026-04-28
Exercise all three feature modes in placement shim e2e tests via header override

PhilippMatthes:
The middleware now reads X-Cortex-Feature-Mode from every request and, when valid, stores it in context. featureModeFromConfOrHeader resolves the effective mode per handler (override if present, configured default otherwise). All handlers use this helper instead of reading config directly.

Each e2e test is wrapped with e2eWrapWithModes which iterates passthrough, hybrid, and crd. A custom RoundTripper (e2eModeTransport) auto-injects the header from context so individual tests need no changes. Passthrough-only endpoints probe for 501 via e2eProbeUnimplemented and skip gracefully. Tests that require CRD infrastructure (traits CRUD, RP traits writes) gate on the configured mode. The resourceLocker is now always initialized to avoid nil panics from mode overrides.

### DIFF
--- a/internal/shim/placement/handle_allocation_candidates_e2e.go
+++ b/internal/shim/placement/handle_allocation_candidates_e2e.go
@@ -47,6 +47,15 @@ func e2eTestAllocationCandidates(ctx context.Context, _ client.Client) error {
 	const testRC = "CUSTOM_CORTEX_E2E_CAND_RC"
 	const apiVersion = "placement 1.26"
 
+	// Probe: for non-passthrough modes, verify endpoint returns 501.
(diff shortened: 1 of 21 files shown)


PhilippMatthes:
Addressed both review comments in 3f4e3b8c:

1. **Guard against missing backing config** (shim.go): `featureModeFromConfOrHeader` now rejects overrides to hybrid/crd when neither `Versioning` nor `Traits` config is present, falling back to the configured default. This prevents nil-pointer dereferences in handlers that depend on startup-validated config. Added unit tests covering this guard.

2. **Improve probe error handling** (shim_e2e.go): `e2eProbeUnimplemented` now returns an error for unexpected 4xx/5xx status codes (other than 501) instead of silently treating them as "endpoint works". Only 2xx responses cause `(false, nil)` to be returned.

PR #772, opened 2026-04-29
Feature-gated /resource_classes API with ConfigMap storage

PhilippMatthes:
Implements three-mode support for the /resource_classes placement API endpoints, following the same pattern established in the traits refactor (PR #771). Previously these endpoints only forwarded to upstream placement (returning 501 for hybrid/crd modes). Now passthrough forwards to upstream, hybrid forwards and mirrors locally, and crd serves entirely from a shim-owned ConfigMap. A ResourceClassSyncer periodically fetches upstream state into the ConfigMap so that crd mode can serve without depending on live upstream availability. This is part of the phased placement API migration.

Assisted-by: claude-code:claude-opus-latest [Bash] [Read]

### DIFF
--- a/helm/bundles/cortex-placement-shim/values.yaml
+++ b/helm/bundles/cortex-placement-shim/values.yaml
@@ -62,6 +62,8 @@ cortex-shim:
       status: "CURRENT"
     traits:
       configMapName: "cortex-placement-shim-traits"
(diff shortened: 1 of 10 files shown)


PhilippMatthes:
I'll wait so coderabbit has a chance to review

PR #773, opened 2026-04-29
feat: Controllers and http API use CR CRD for managing CR reservation CRDs

mblos:
Introduces the CommittedResource CRD as the durable contract between the LIQUID API and the reservation system, replacing direct Reservation writes with a controller-owned lifecycle. Accepting or rejecting a commitment is now a write to a CommittedResource CRD followed by watching its status — the API layer no longer touches Reservation CRDs directly.


mblos:
@coderabbitai full review

PR #787, opened 2026-05-04
Implement placement shim allocations via hv1 bookings

PhilippMatthes:
The Cortex Placement Shim needs to serve the Placement API's allocations endpoints (`GET/PUT/DELETE /allocations/{consumer_uuid}` and `POST /allocations`) directly from Hypervisor CRDs, without always forwarding to upstream Placement. This PR implements those handlers by reading and writing `spec.bookings` on Hypervisor CRs — the field introduced in cobaltcore-dev/openstack-hypervisor-operator#296.

Three operating modes are supported per request, configured globally or overridden via request header:

- **Passthrough** — forward the request to upstream Nova Placement unchanged.
- **Hybrid** — split allocations by resource provider: KVM-managed providers (those that resolve to a Hypervisor CR via `status.hypervisorId`) are written to CRD bookings; all others are forwarded upstream. GET merges both sources; DELETE removes from both.
- **CRD** — serve entirely from Hypervisor CRDs. All resource providers in the request must resolve to a known CR.

Key implementation details:

- A new field index (`spec.bookings.consumer.uuid`) enables fast lookup of all Hypervisors that hold bookings for a given consumer UUID, including the multi-host case during live migrations.
- Resource translation between Placement resource classes (`VCPU`, `MEMORY_MB`) and CRD quantities (`cpu`, `memory` as Kubernetes `resource.Quantity`) is handled by `placementToHVResources` / `hvToPlacementResources`.
- `consumer_generation` concurrency control is enforced: null generation means new consumer, non-null must match the stored generation, mismatches return 409 Conflict.
- `POST /allocations` (batch manage, microversion 1.13+) supports atomic multi-consumer writes — the primary mechanism for live-migration move operations.
- A `statusRecorder` captures upstream responses in hybrid mode so the shim can inspect the status code before deciding whether to proceed with CRD writes.

The `go.mod` dependency is bumped to pull in the `spec.bookings` types and helper functions (`GetConsumers`, `GetConsumer`, `HasConsumer`) from the operator.

Tests cover: unit tests for resource translation, per-mode unit tests for all four HTTP verbs, hybrid-mode merge and upstream-failure scenarios, and a full e2e test suite exercising both passthrough (with a dynamically created RP and custom resource class) and CRD paths against a real cluster.

See: https://github.com/cobaltcore-dev/openstack-hypervisor-operator/pull/296

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/cobaltcore-dev/cortex
 go 1.26.0
 
 require (
-	github.com/cobaltcore-dev/openstack-hypervisor-operator v1.2.0
+	github.com/cobaltcore-dev/openstack-hypervisor-operator v1.1.1-0.20260430095528-58fbe3ff4c3e
 	github.com/go-gorp/gorp v2.2.0+incompatible
 	github.com/gophercloud/gophercloud/v2 v2.12.0
 	github.com/ironcore-dev/ironcore v0.3.0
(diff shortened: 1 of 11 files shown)


PhilippMatthes:
Superceded by internal ADR making the vm crd the source of truth for shim allocations.

PR #793, opened 2026-05-04
Release minor fixes - Monday May 4 2026

PhilippMatthes:
Change summary:

### cortex v0.0.46 (sha-ab6eb45d)

Non-breaking changes:
- Fix capacity filter to correctly account for multi-VM CommittedResource reservation slots — confirmed VMs are now summed (not just the last one), blocks are clamped to zero when confirmed exceeds slot size, and spec-only VMs larger than remaining slot are fully covered
- Expose `prometheusDatasourceControllerParallelReconciles` config option to allow parallel reconciles in the Prometheus datasource controller, reducing initial sync latency
- Remove `Conf` field from PrometheusDatasourceReconciler — config is now loaded internally via `conf.GetConfig` during `SetupWithManager`
- Add operator-controlled per-resource-type config (`flavorGroupResourceConfig`) for committed resources, replacing runtime derivation from flavor group metadata; supports wildcard (`*`) catch-all for unknown groups
- Propagate `AnnotationCreatorRequestID` from the change-commitments API to the CommittedResource CRD and through the reservation controller for end-to-end request tracing

### cortex-nova v0.0.59 (sha-ab6eb45d)

Includes updated chart cortex v0.0.46.

Non-breaking changes:
- Remove all committed resource related Prometheus alerts (info API, change API, usage API, capacity API, and syncer alerts)
- Add `flavorGroupResourceConfig` to cortex-nova values.yaml with a wildcard default that sets `hasCapacity: true` for ram, cores, and instances


PhilippMatthes:
@claude run the /release 793 command

PhilippMatthes:
@claude run the /release 793 command again since there are new changes. Also make sure the version bump PR is opened.

PhilippMatthes:
@claude in the opened bump-pr, please also perform a patch-bump of the library chart and the bundle charts depending on this library chart, and then adjust the changelog with the correct version identifiers of the respective charts.

umswmayj (approved):

juliusclausnitzer (approved):

PR #796, opened 2026-05-04
feat: Qutoa add CRD and calc payg

umswmayj:


umswmayj:
follow up of https://github.com/cobaltcore-dev/cortex/pull/783

PR #799, opened 2026-05-05
feat: adding pipeline options

mblos:
Introduce call-time pipeline options as the foundation for consolidating duplicate pipelines — behavioral variations (lock reservations, record history, pessimistic blocking) become explicit flags instead of separate pipelines


internal/scheduling/lib/filter_monitor.go
```diff
@@ -43,6 +43,6 @@ func (fm *FilterMonitor[RequestType]) Validate(ctx context.Context, params v1alp
 }
 
 // Run the filter and observe its execution.
-func (fm *FilterMonitor[RequestType]) Run(traceLog *slog.Logger, request RequestType) (*FilterWeigherPipelineStepResult, error) {
-	return fm.monitor.RunWrapped(traceLog, request, fm.filter)
+func (fm *FilterMonitor[RequestType]) Run(traceLog *slog.Logger, request RequestType, opts Options) (*FilterWeigherPipelineStepResult, error) {
```
PhilippMatthes: `request RequestType` is constrained by `type FilterWeigherPipelineRequest interface` which now carries the new `GetOptions()` method. So we should remove all additional `opts Options` parameters.

internal/scheduling/lib/filter_weigher_pipeline.go
```diff
@@ -279,7 +286,7 @@ func (p *filterWeigherPipeline[RequestType]) Run(request RequestType) (v1alpha1.
 
 	// Run filters first to reduce the number of hosts.
 	// Any weights assigned to filtered out hosts are ignored.
-	filteredRequest, filterStepResults := p.runFilters(traceLog, request)
+	filteredRequest, filterStepResults := p.runFilters(traceLog, request, opts)
```
PhilippMatthes: Please revert this change. Scheduler steps can get the options from the provided request.

internal/scheduling/lib/filter_weigher_pipeline.go
```diff
@@ -290,13 +297,23 @@ func (p *filterWeigherPipeline[RequestType]) Run(request RequestType) (v1alpha1.
...
+	if opts.MaxCandidates > 0 && len(hosts) > opts.MaxCandidates {
+		hosts = hosts[:opts.MaxCandidates]
+		// Drop trimmed hosts from outWeights so AggregatedOutWeights stays consistent.
+		for host := range outWeights {
+			if !slices.Contains(hosts, host) {
+				delete(outWeights, host)
+			}
+		}
+	}
+
```
PhilippMatthes, line 332: Please provide logging here so we see what's going on.

internal/scheduling/lib/options.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/cobaltcore-dev/cortex/api/v1alpha1"
+)
+
+// Options configure the behavior of a single pipeline run at call time.
+// These are distinct from per-step YAML options (FilterWeigherPipelineStepOpts),
+// which are static and set when the pipeline is initialized.
+//
+// Consumed by steps: ReadOnly, LockReservations, AssumeEmptyHosts, IgnoredReservationTypes.
+// Consumed by the controller after pipeline.Run(): RecordHistory, CreateInflight.
```
PhilippMatthes: These two code comment lines are likely to be obsolete once the controller or step logic changes. We should consider removing them.

api/scheduling/options.go
```diff
@@ -0,0 +1,49 @@
...
+// Options configure the behavior of a single pipeline run at call time.
+// These are distinct from per-step YAML options (FilterWeigherPipelineStepOpts),
+// which are static and set when the pipeline is initialized.
+//
+// Consumed by steps: ReadOnly, LockReservations, AssumeEmptyHosts, IgnoredReservationTypes.
+// Consumed by the controller after pipeline.Run(): RecordHistory, CreateInflight.
+type Options struct {
+	// ReadOnly means the pipeline run does not modify shared scheduling state (reservations,
+	// history, inflight records). Concurrent read-only runs are safe under a shared read lock.
+	// Note: the controller may still write the Decision status after Run() regardless of this flag.
```
PhilippMatthes: This seems inconsistent. Shouldn't we draw the line here? Read-only requests create or modify NO resources and are purely to calculate host candidates for constraints.

internal/scheduling/lib/options.go
```diff
@@ -0,0 +1,49 @@
...
+type Options struct {
+	// ReadOnly means the pipeline run does not modify shared scheduling state (reservations,
+	// history, inflight records). Concurrent read-only runs are safe under a shared read lock.
+	// Note: the controller may still write the Decision status after Run() regardless of this flag.
+	ReadOnly bool
+	// LockReservations prevents reservation unlocking, e.g. in the capacity filter.
+	// Set when finding hosts for new reservations (failover, CR) to see true available capacity.
+	LockReservations bool
```
PhilippMatthes: Should this be more generic such as `Kind` which is a typed enum? In this case, the capacity filter would just check for `req.GetOptions().Kind == KindFailoverReservation` to control which logic is executed. We could also add a kind `KindCapacityScan` for limes etc. -- this is nicely extensible and well-defined. In this case, the `ReadOnly`, `AssumeEmptyHosts`, and `CreateInFlight` flags could also be removed.

internal/scheduling/lib/options.go
```diff
@@ -0,0 +1,49 @@
...
+	// AssumeEmptyHosts treats all hosts as having no running VMs.
+	AssumeEmptyHosts bool
+	// IgnoredReservationTypes lists reservation types the capacity filter skips entirely.
+	IgnoredReservationTypes []v1alpha1.ReservationType
```
PhilippMatthes: Should we make this a substruct such as `type ReservationOptions struct`?

PhilippMatthes (changes requested):
I like the idea of extending the external scheduler request(s) with an options struct. However, we should go over the fields exposed by this struct once again. I am not sure about putting it under `internal/scheduling/lib` as we have lots of fields that will only be used by the Nova integration of cortex. So, why not put the options struct under the nova external scheduler request struct? In that way, you don't need to extend the `lib` interface and conform all other packages.

internal/scheduling/lib/options.go
mblos: Not sure I get this comment, probably better to discuss live

api/scheduling/options.go
mblos: There are use cases, we can discuss live

mblos:
@coderabbitai full review

PhilippMatthes (approved):
lgtm, thanks for considering my feedback

PR #826, opened 2026-05-08
feat: adding CPU committed resources

mblos:
Extends committed resource support from RAM-only to include CPU core commitments.


mblos:
@coderabbitai full review

PR #834, opened 2026-05-11
ci: remove non docker tests

umswmayj:


mblos (approved):

SoWieMarkus (commented):
The purpose of test-without-docker is to act as a fast-fail gate by running the tests without spinning up any docker containers and catches obvious failures quickly (and cheaply) before the heavier test-with-docker job is triggered.

I don't have a problem with rremoving this gate. But if we do, the test-without-docker job becomes redundant, so we could drop it entirely. We just need to scope the `test-with-docker` job to cover the whole repository and not just the `./internal/...` directory.

PR #847, opened 2026-05-12
feat: track VM placements in reservations and classify no-host-found

mblos:
Adds two pieces of observability for committed-resource scheduling outcomes.

- After a successful placement, the placed VM is recorded in the reservation slot so the scheduler knows which capacity is actually occupied
- When placement returns no host for a CR flavor, the failure is classified by root cause: no active commitment, commitment exhausted, no free reservation slot, or capacity available but filtered out
- Only pipelines with history enabled participate, keeping probe pipelines noise-free


SoWieMarkus (dismissed):

SoWieMarkus:
Let me check why that alerts workflow is failing

SoWieMarkus (dismissed):

internal/scheduling/nova/features.go
```diff
@@ -0,0 +1,19 @@
...
+	// CommittedResourceTracking enables committed resource integration: writing
+	// placed VM UUIDs back into Reservation slots and classifying no-host-found
+	// results by committed resource coverage. Enable only on deployments that
+	// use committed resources.
+	CommittedResourceTracking bool `json:"committedResourceTracking,omitempty"`
+	// PessimisticBlocking enables blocking all candidate hosts at scheduling time
+	// to prevent concurrent placements from violating capacity constraints. Enable
+	// only on deployments that support pessimistic blocking.
+	PessimisticBlocking bool `json:"pessimisticBlocking,omitempty"`
+}
```
PhilippMatthes, line 15: I feel like this should be in the internal package that needs it (e.g. internal/scheduling/nova)

mblos: actually `PessimisticBlocking` was just to show what kind of gates we could add. I will remove it here, its not needed anyway. 
No strong opinion on having the feature gates globally or domain specific, will move them to nova for now

PhilippMatthes: If we want to have it globally, we should put it under internal/scheduling not under pkg. At the moment this is not something we offer as a library someone else can build own go code on (what `pkg` is for).

PhilippMatthes: Thanks for moving it!

internal/scheduling/nova/cr_allocation.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/cobaltcore-dev/cortex/internal/knowledge/extractor/plugins/compute"
+	"github.com/cobaltcore-dev/cortex/internal/scheduling/reservations"
+	hv1 "github.com/cobaltcore-dev/openstack-hypervisor-operator/api/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// recordCRAllocation writes the placed VM UUID into the matching Reservation
+// Spec.CommittedResourceReservation.Allocations after a successful Nova placement.
+func (c *FilterWeigherPipelineController) recordCRAllocation(ctx context.Context, decision *v1alpha1.Decision, request api.ExternalSchedulerRequest) {
```
PhilippMatthes: This functionality can be isolated into its own subpackage `internal/scheduling/nova/crs`. I think this will help us in the long run building out the code in a clear structure here.

You could do a `crs.Recorder` and embed this struct into the `FilterWeigherPipelineController`:

```go
type FilterWeigherPipelineController struct {
	// Toolbox shared between all pipeline controllers.
	lib.BasePipelineController[lib.FilterWeigherPipeline[api.ExternalSchedulerRequest]]
  // ...
  // crRecorder receives scheduling events
  // and updates reservations appropriately.
  crRecorder crs.Recorder
}

// package internal/scheduling/nova/crs
type Config struct { /* ... */ }

// package internal/scheduling/nova/crs
type Recorder struct {
  config Config

  // NoHostFoundCounter counts no-host-found results by CR slot outcome (no_cr/cr_exhausted/slot_exhausted/slot_blocked/error).
	NoHostFoundCounter *prometheus.CounterVec
	// PlacementCounter counts successful placements by CR slot outcome.
	PlacementCounter *prometheus.CounterVec
}

func (r *Recorder) Record(ctx context.Context, decision *v1alpha1.Decision, request api.ExternalSchedulerRequest) { /* ... */ }
```

auhlig (dismissed):
Cool!

internal/scheduling/nova/cr_allocation.go
mblos: I like the proposal, tasked chatty do move the code

internal/scheduling/nova/filter_weigher_pipeline_controller.go
```diff
@@ -41,10 +42,15 @@ type FilterWeigherPipelineController struct {
...
 
+	// FeatureGates holds feature flags for this controller.
+	FeatureGates FeatureGates
 	// Monitor to pass down to all pipelines.
 	Monitor lib.FilterWeigherPipelineMonitor
 	// Candidate gatherer to get all placement candidates if needed.
 	gatherer CandidateGatherer
+
+	// CRRecorder receives scheduling events and updates CR reservations and metrics.
+	CRRecorder crs.Recorder
```
PhilippMatthes, line 53: Nit: do we need to export this field to other packages?

PhilippMatthes (approved):
Thanks for considering my feedback!

internal/scheduling/nova/filter_weigher_pipeline_controller.go
mblos: hm, its for setting the multicluster client similar to how we do it with others.. so I sticked to the pattern we have so far. if no strong opinion, I would keep it as it is

umswmayj (approved):

PR #922, opened 2026-06-08
feat: use replace oci path script for workflows

SoWieMarkus:


.github/workflows/check-alerts.yaml
```diff
@@ -22,12 +22,8 @@ jobs:
           set -euo pipefail
           mkdir -p rendered
 
-          # Rewrite library chart dependencies to local file paths so helm dep
-          # update does not try to pull them from OCI (they are not published).
           for bundle in helm/bundles/cortex-cinder helm/bundles/cortex-manila helm/bundles/cortex-nova helm/bundles/cortex-placement-shim; do
```
PhilippMatthes: Can we use the changed files workflow to populate this?

PR #929, opened 2026-06-10
fix(renovate): pin python to 3.13 and harden config rules

umswmayj:
- Revert python-version to 3.13 in setup-claude-code-action/action.yml for compatibility with upstream anthropics/claude-code-action.
- Revert constraints.python to 3.13 in renovate.json.
- Replace broken 'matchManagers: [tool-constraint]' rule (which never matched anything; tool-constraint is a depType, not a manager id) with 'matchDepTypes: [tool-constraint], enabled: false'. This is the actual root cause that allowed renovate to bump constraints.python from 3.13 to 3.14 in #908.
- Add 'matchPackageNames: [python], allowedVersions: 3.13.x' as a hard version cap.
- Add file-scoped 'automerge: false' rule for .github/actions/setup-claude-code-action/action.yml as a tripwire so any future bump in this file requires manual review.

Rules are placed after the 'External dependencies' catch-all so that later-rule-wins semantics override the catch-all's automerge: true.

### DIFF
--- a/.github/actions/setup-claude-code-action/action.yml
+++ b/.github/actions/setup-claude-code-action/action.yml
@@ -41,7 +41,7 @@ runs:
     - name: Setup Python
       uses: actions/setup-python@v6
       with:
-        python-version: "3.14"
+        python-version: "3.13"
 
     - name: Install LiteLLM dependencies
       shell: bash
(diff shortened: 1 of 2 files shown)


PhilippMatthes (approved):
Let's see

PR #940, opened 2026-06-15
ci(renovate): track Helm chart appVersions via custom regex managers

umswmayj:
Add three custom regex managers so Renovate detects new sha-* image tags on ghcr.io and opens PRs bumping the appVersion in the matching Chart.yaml files (cortex, cortex-shim, cortex-postgres).

Replaces the direct-push update-appversion.yml workflow, which is broken by branch protection. The old workflow is kept as-is for now.

Automerge is disabled for these PRs initially so the behavior can be verified before enabling auto-merge.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -69,6 +69,36 @@
       "datasourceTemplate": "docker",
       "versioningTemplate": "semver-coerced",
       "autoReplaceStringTemplate": "ENV PG_VERSION {{{newValue}}}-1.pgdg13+1"
+    },
+    {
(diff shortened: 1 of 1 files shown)


SoWieMarkus:
Nice idea! Lets say: Better than nothing. 

But: From my experience Renovate takes ages to open a PR in this repository when you trigger one manually. So I can imagine it will take 15min+ for that PR to open up. In that time I have already bumped the commit sha myself to see my changes in staging :D

SoWieMarkus (approved):

PR #942, opened 2026-06-15
feat(capacity): improve FlavorGroupCapacity CRD observability

juliusclausnitzer:


internal/scheduling/reservations/capacity/metrics.go
```diff
@@ -138,6 +146,15 @@ func (m *Monitor) Collect(ch chan<- prometheus.Metric) {
 		}
 		m.exclusivelyFreeSlots.With(groupAZFlavorLabels).Set(float64(crd.Status.ExclusivelyFreeSlots))
 
+		readyVal := 0.0
+		for _, cond := range crd.Status.Conditions {
```
mblos: I suggest to use `apimeta.IsStatusConditionTrue(crd.Status.Conditions, v1alpha1.FlavorGroupCapacityConditionReady)`

mblos (changes requested):
Thanks for the PR 👍 

- we should update the capacity controller in this pr.. it never writes Ready=False on failure.
- once the controller sets ready=false, the capacity API should return 5xx

internal/scheduling/reservations/commitments/capacity.go
```diff
@@ -86,6 +92,7 @@ func (c *CapacityCalculator) CalculateCapacity(ctx context.Context, req liquid.S
 			if !apimeta.IsStatusConditionTrue(crd.Status.Conditions, v1alpha1.FlavorGroupCapacityConditionReady) {
 				logger.Info("FlavorGroupCapacity CRD is stale, reporting capacity without usage",
```
mblos: outdated

internal/scheduling/reservations/commitments/api/report_capacity.go
```diff
@@ -77,7 +78,11 @@ func (api *HTTPAPI) HandleReportCapacity(w http.ResponseWriter, r *http.Request)
 	report, err := calculator.CalculateCapacity(ctx, req)
 	if err != nil {
 		logger.Error(err, "failed to calculate capacity")
```
mblos: always log.error?
log should distinguish between 

- one or more FlavorGroupCapacity CRDs are not ready: flavorGroup=X ..
- failed to calculate capacity

mblos (dismissed):
small things left

mblos (approved):

PR #950, opened 2026-06-17
feat(scheduling): migrate CR and failover calls to options-based pipeline selection

mblos:
Eliminating dedicated CR and failover pipelines by encoding their behavioral differences as call-time scheduling options, so the standard GP and HANA pipelines are sufficient for all internal scheduling calls.

- Dedicated pipelines (kvm-report-capacity, kvm-*-no-history, kvm-valid-host-reuse-failover-reservation, kvm-acknowledge-failover-reservation) are no longer needed — their behavioral differences are now expressed via call-time options, not separate pipeline CRDs
- Capacity probes and failover scheduling now select GP vs HANA pipeline based on the VM/flavor's traits
- Default helm values updated; deployments referencing dedicated pipelines can be updated later on


mblos:
as of now, tests are redundant. no strong opinion on either joint testing across filters like done in  `filter_skip_placement_context_test.go`  .. or separated in individual files. tbd

mblos:
changed semantic of failover calls.. now selecting gp/hana pipelines. Please check if we want to do it like this or always go with gp pipeline

internal/scheduling/reservations/capacity/controller.go
```diff
@@ -291,17 +291,37 @@ func (c *Controller) probeScheduler(
...
 		MemoryMB:         flavor.MemoryMB,
 		VCPUs:            flavor.VCPUs,
 		FlavorExtraSpecs: flavor.ExtraSpecs,
 		AvailabilityZone: az,
 		Pipeline:         pipeline,
 		EligibleHosts:    eligibleHosts,
-	}, scheduling.Options{SkipHistory: true, SkipInflight: true, SkipCommittedResourceTracking: true})
+	}, scheduling.Options{
+		ReadOnly:                      true,
+		AssumeEmptyHosts:              ignoreAllocations,
```
juliusclausnitzer, line 555: If `AssumeEmptyHosts` and `ignoreAllocations` are synonymous, we should probably stick with one and get rid of the other

mblos: Yes, they're synonymous — IgnoreAllocations in YAML and AssumeEmptyHosts in Options both do the same thing, merged with OR in the filter. IgnoreAllocations can be removed once kvm-report-capacity is deleted 
We keep all the dedicated pipelines for now for migration purposes..

juliusclausnitzer (dismissed):

internal/scheduling/reservations/failover/reservation_scheduling.go
```diff
@@ -123,7 +123,7 @@ func (c *FailoverReservationController) tryReuseExistingReservation(
 
 	logger := LoggerFromContext(ctx)
 
-	validHypervisors, err := c.queryHypervisorsFromScheduler(ctx, vm, allHypervisors, PipelineReuseFailoverReservation, resSpec, scheduling.Options{ReadOnly: true, SkipHistory: true, SkipInflight: true, SkipCommittedResourceTracking: true})
+	validHypervisors, err := c.queryHypervisorsFromScheduler(ctx, vm, allHypervisors, inferFailoverPipeline(vm.FlavorExtraSpecs), resSpec, scheduling.Options{ReadOnly: true, SkipPlacementContextFilters: true, SkipHistory: true, SkipInflight: true, SkipCommittedResourceTracking: true})
```
umswmayj: Not sure if this is correct.
PipelineReuseFailoverReservation skips capacity related checks like: `has_enough_capacity` or `quota` but still needs to do e.g., `anti affinity` (not sure if affinity is reasonable during failover). Looking at the yaml the `anti-affinity` is missing currently. I think that was not yet active and is a good point why we do not want duplicated pipelines.

internal/scheduling/reservations/failover/reservation_scheduling.go
```diff
@@ -222,9 +222,9 @@ func (c *FailoverReservationController) validateVMViaSchedulerEvacuation(
 		"vmUUID", vm.UUID,
 		"reservationHost", reservationHost,
 		"vmCurrentHost", vm.CurrentHypervisor,
-		"pipeline", PipelineAcknowledgeFailoverReservation)
+		"pipeline", scheduleReq.Pipeline)
 
-	resp, err := c.SchedulerClient.ScheduleReservation(ctx, scheduleReq, scheduling.Options{ReadOnly: true, LockReservations: true, SkipHistory: true, SkipInflight: true, SkipCommittedResourceTracking: true})
+	resp, err := c.SchedulerClient.ScheduleReservation(ctx, scheduleReq, scheduling.Options{ReadOnly: true, LockReservations: true, SkipPlacementContextFilters: true, SkipHistory: true, SkipInflight: true, SkipCommittedResourceTracking: true})
```
umswmayj: Again not sure about `SkipPlacementContextFilters: true`.

internal/scheduling/nova/plugins/filters/filter_has_enough_capacity.go
```diff
@@ -173,14 +173,23 @@ func (s *FilterHasEnoughCapacity) Run(traceLog *slog.Logger, request api.Externa
...
+					case api.EvacuateIntent:
+						if reservation.Status.FailoverReservation != nil {
+							if _, contained := reservation.Status.FailoverReservation.Allocations[request.Spec.Data.InstanceUUID]; contained {
+								traceLog.Info("unlocking resources reserved by failover reservation for VM in allocations (evacuation)",
+									"reservation", reservation.Name,
+									"instanceUUID", request.Spec.Data.InstanceUUID)
+								continue
+							}
 						}
+					case api.ReuseFailoverReservationIntent:
```
umswmayj, line 187: Should be OK, minor semantic change: the old pipeline skipped the entire capacity check. But as we have a fitting failover reservation, this should be OK.

internal/scheduling/reservations/failover/reservation_scheduling.go
umswmayj: bump

internal/scheduling/reservations/failover/reservation_scheduling.go
```diff
@@ -266,7 +272,7 @@ func (c *FailoverReservationController) scheduleAndBuildNewFailoverReservation(
 
 	// Get potential hypervisors from scheduler using the reservation spec resources
 	// (which may be sized to the LargestFlavor from the flavor group)
-	validHypervisors, err := c.queryHypervisorsFromScheduler(ctx, vm, allHypervisors, PipelineNewFailoverReservation, resSpec, scheduling.Options{LockReservations: true, SkipHistory: true, SkipInflight: true, SkipCommittedResourceTracking: true})
+	validHypervisors, err := c.queryHypervisorsFromScheduler(ctx, vm, allHypervisors, PipelineNewFailoverReservation, resSpec, api.ReserveForFailoverIntent, scheduling.Options{LockReservations: true, SkipPlacementContextFilters: false, SkipHistory: true, SkipInflight: true, SkipCommittedResourceTracking: true})
```
umswmayj: Can we add a `DefaultFailoverOptions` and use that for all other calls and here then:

```
validHyperOptions := DefaultFailoverOptions
validHyperOptions.LockReservations = True
validHypervisors, err := c.queryHypervisorsFromScheduler(ctx, vm, allHypervisors, PipelineNewFailoverReservation, resSpec, api.ReserveForFailoverIntent, validHyperOptions)
```

If i am not mistaken this is the only call where we change a value

mblos: Good point, done.

internal/scheduling/nova/plugins/filters/filter_has_enough_capacity.go
mblos: Agreed

internal/scheduling/reservations/failover/reservation_scheduling.go
mblos: good catch, fixed

umswmayj (dismissed):

internal/scheduling/nova/plugins/filters/filter_aggregate_metadata.go
```diff
@@ -21,7 +21,10 @@ type FilterAggregateMetadata struct {
 // the "filter_tenant_id" metadata key set.
 func (s *FilterAggregateMetadata) Run(traceLog *slog.Logger, request api.ExternalSchedulerRequest) (*lib.FilterWeigherPipelineStepResult, error) {
 	result := s.IncludeAllHostsFromRequest(request)
-	if request.GetOptions().SkipPlacementContextFilters {
+	// Failover and capacity probe calls are not placed on behalf of a tenant project;
+	if intent, err := request.GetIntent(); err == nil &&
+		(intent == api.ReserveForFailoverIntent || intent == api.ReuseFailoverReservationIntent ||
+			intent == api.CapacityProbeIntent) {
```
PhilippMatthes: Only nitpick: I'd use slices.Contains here

PhilippMatthes (dismissed):
Thanks for addressing my idea! Looks good, nice job.

PhilippMatthes (approved):

PR #951, opened 2026-06-17
feat(cr): pre-allocate PAYG VMs into CR reservation slots on cr creation/modification

mblos:
When a new committed resource is created, the controller now absorbs existing PAYG VMs into reservation slots instead of blindly probing the scheduler for every slot.


internal/scheduling/reservations/commitments/reservation_manager.go
```diff
@@ -194,16 +204,78 @@ func (m *ReservationManager) ApplyCommitmentState(
...
+				allCandidates = append(allCandidates, hvCandidates...)
+			}
+			sortCandidatesDesc(allCandidates)
+
+			for _, candidate := range allCandidates {
+				if deltaMemoryBytes <= 0 {
+					break
+				}
+				slotMemoryMB := candidate.MemoryMB
+				if int64(slotMemoryMB)*1024*1024 > deltaMemoryBytes { //nolint:gosec
```
juliusclausnitzer: in theory, should we not have conversion methods for this instead of using *1024? 

Also, if we have one 512GB Vm and one 256GB VM and 256GB deltaMemory, why would we rather partially consume the big VM instead of taking the fitting slot?

mblos: agree with the 1024 conversion... but should be done in a separated pr. there are many places n the code base

juliusclausnitzer (approved):

SoWieMarkus (approved):

PR #1015, opened 2026-07-06
feat: mutex-protected in-process map for scheduling safety

SoWieMarkus:
Controller-runtime's informer is eventually consistent - a write via the client is not immediately visible to subsequent reads within the same pod. This PR adds a package that wraps the multicluster client and keeps a short-lived in-process map of recently written objects so reads see them immediately.

- CRDCache holds assumed objects keyed by namespace/name with TTL-based eviction
- CachingClient wraps client.Client and transparently populates the cache on writes, overlays cached entries on reads
- An informer-based eviction handler evicts entries as soon as the informer catches up
- Which GVKs get cached and the TTL are configurable via `conf.crdCache` in values.yaml; when no GVKs are listed, all writes are cached
- The caching client is now wired everywhere in `main`. Controllers that don't benefit from caching simply pass through

One open question where I would like your feedback: should the TTL timer be replaced with a timestamp-per-entry approach? A timer fires from a separate goroutine and can race with a concurrent Assume for the same key -Stop() returning false means the goroutine is already queued and will evict the freshly assumed entry once it acquires the lock. A timestamp checked lazily in Get/List avoids this entirely, at the cost of not proactively freeing memory (entries idle around until next read of this item). We could also tackle that with an additional periodical clean up task.


pkg/crdcache/client.go
```diff
@@ -0,0 +1,142 @@
...
+// Update delegates to the inner client and, on success, updates the cache entry.
+func (c *CachingClient) Update(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
+	if err := c.Client.Update(ctx, obj, opts...); err != nil {
+		return err
+	}
+	if c.shouldCache(obj) {
+		c.cache.Assume(obj)
+	}
+	return nil
+}
```
PhilippMatthes, line 57: What happens when there are two concurrent calls to this function?
Can there be a race condition w.r.t the client update executing in another order than the cache update? Ideally, we should add a globally shared mutex across all CRUD functions so that concurrently reading/writing routines see a consistent state.

pkg/crdcache/eviction.go
```diff
@@ -0,0 +1,48 @@
...
+			}
+		},
+		DeleteFunc: func(obj any) {
+			if o, ok := obj.(client.Object); ok {
+				c.Forget(ObjectKey(o))
+			}
+		},
+	})
+	return err
+}
```
PhilippMatthes, line 48: Any specific reason why this is outside `cache.go`?

pkg/crdcache/options.go
```diff
@@ -0,0 +1,108 @@
...
+			if _, ok := allowed[gvk]; ok {
+				return true
+			}
+		}
+		return false
+	}
+}
+
+// Duration is a time.Duration that marshals/unmarshals as a Go duration string (e.g. "2m30s").
+type Duration struct{ time.Duration }
```
PhilippMatthes, line 77: Is it possible to reuse `metav1.Duration` here?

pkg/crdcache/client.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Get checks the cache first; on miss it delegates to the inner client.
+func (c *CachingClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	cacheKey := key.Namespace + "/" + key.Name
+	if cached, ok := c.cache.Get(cacheKey); ok {
+		reflect.ValueOf(obj).Elem().Set(reflect.ValueOf(cached).Elem())
+		return nil
+	}
+	return c.Client.Get(ctx, key, obj, opts...)
```
PhilippMatthes, line 86: What if the resource has already been deleted from the kubernetes apiserver? Doesn't this mean that for up to 1x TTL your controller works on an inconsistent state? Sounds nasty

pkg/crdcache/client.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// List calls the inner client and overlays cached entries not yet visible in the informer result.
+// Objects already present in the informer result are not replaced — informer truth takes precedence.
+// Objects present in the cache but absent from the informer are appended to the result.
+func (c *CachingClient) List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+	if err := c.Client.List(ctx, list, opts...); err != nil {
+		return err
+	}
```
PhilippMatthes, line 95: This'll work for created resources (they will be in the returned list), but will it also work for deleted or updated items?

pkg/crdcache/client_test.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func mustParseGVK(t *testing.T, s string) GVK {
+	t.Helper()
+	var g GVK
+	if err := g.UnmarshalJSON([]byte(`"` + s + `"`)); err != nil {
+		t.Fatalf("parse GVK %q: %v", s, err)
+	}
+	return g
+}
```
PhilippMatthes, line 294: I'd love to see an e2e test that validates the following:
- Without the caching client, we see outdated items sharing across 2 goroutines
- With the caching client, we don't see outdated items sharing across 2 goroutines

pkg/crdcache/options.go
```diff
@@ -0,0 +1,108 @@
...
+// Config is the top-level values.yaml-configurable struct for CRDCache.
+type Config struct {
+	CRDCache CRDCacheSettings `json:"crdCache"`
+}
+
+// CRDCacheSettings holds the per-cache settings nested under the crdCache key.
+type CRDCacheSettings struct {
+	// TTL is the maximum time an assumed entry stays in the cache before being
+	// evicted by the TTL timer. Defaults to DefaultTTL when zero.
+	TTL Duration `json:"ttl"`
```
PhilippMatthes, line 29: Can't we use the informer-cache's callbacks (e.g. `handler.EnqueueRequestForOwner`) to keep track of resource changes from the upstream client and drain outdated resources from the cache synchronously?

PhilippMatthes (changes requested):
Love it! Anyhow, there are probably some things we want to consider.

pkg/crdcache/client.go
PhilippMatthes: See https://github.com/cobaltcore-dev/cortex/pull/1015#discussion_r3528240706

SoWieMarkus:
https://github.com/cobaltcore-dev/cortex/pull/1042

PR #1024, opened 2026-07-08
refactor: number of synced objects

SoWieMarkus:
- Change `NumberOfObjects` from int64 to *int64 so omitempty no longer strips the field when zero, making `kubectl get datasource` display 0 instead of blank
- Update the skip-sync guard to != nil (was != 0) so a successful sync returning 0 objects no longer forces an immediate re-sync
- Scope the `SyncObjectsDroppedToZero` alerts to only fire when the metric was previously non-zero (offset 1h > 0), avoiding false alerts for new datasources/regions


juliusclausnitzer (dismissed):

PhilippMatthes (approved):
Please check if the yaml scaffold needs to be updated and if `omitempty` is still the right choice

PR #1025, opened 2026-07-08
refactor(capacity): convert timer-based controller to controller-runtime reconciler

juliusclausnitzer:
## Summary

The `FlavorGroupCapacity` CRDs were previously updated on a fixed 5-minute timer via a plain goroutine (`Start()` / `manager.RunnableFunc`). This had two problems: the first useful run only happened after the full interval because the Knowledge CRD isn't populated at startup, and there was no reactive trigger for changes to hypervisors, committed resources, or pipelines.

This PR converts the capacity controller to a controller-runtime reconciler:

- **Reactive watches** on Knowledge (flavor_groups only), Hypervisor (capacity/allocation field changes on AZ-labelled hosts), Reservation, and Pipeline — all coalesced to a single reconcile key so rapid bursts of events produce at most one queued reconcile
- **`MinReconcileInterval`** (default 30s) — prevents back-to-back reconciles; early returns with `RequeueAfter: remaining` so no change is ever permanently missed
- **`ReconcileInterval`** (default 5m) acts as the periodic floor, returned as `RequeueAfter` after every successful run
- Config validation enforces `ReconcileInterval > MinReconcileInterval` at startup
- `reconcileAll()` and everything below it is unchanged


internal/scheduling/reservations/capacity/controller.go
```diff
@@ -19,52 +20,111 @@ import (
...
-			timer.Reset(c.config.ReconcileInterval.Duration)
-		}
+// Reconcile implements reconcile.Reconciler. It is called by controller-runtime whenever a
+// watched CRD changes, and also on the periodic RequeueAfter floor set by ReconcileInterval.
+// If called sooner than MinReconcileInterval since the last successful run, it returns early.
+func (c *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
+	c.mu.Lock()
+	elapsed := time.Since(c.lastReconcileAt)
+	minInterval := c.config.MinReconcileInterval.Duration
+	c.mu.Unlock()
```
PhilippMatthes: Can you explain why this mutex is needed? You are setting max concurrent reconciles to 1, so there won't be any race conditions here I assume.

internal/scheduling/reservations/capacity/controller.go
```diff
@@ -19,52 +20,111 @@ import (
...
+
+	if c.lastReconcileAt != (time.Time{}) && elapsed < minInterval {
+		remaining := minInterval - elapsed
+		LoggerFromContext(ctx).V(1).Info("skipping reconcile: min interval not elapsed",
+			"elapsed", elapsed.Round(time.Second),
+			"remaining", remaining.Round(time.Second))
+		return ctrl.Result{RequeueAfter: remaining}, nil
+	}
```
PhilippMatthes: Do I understand this code correctly? You populate the controller-runtime informer cache by watching the corresponding resources (knowledge, hypervisor, committedresource, pipeline) below in `SetupWithManager`. Then, you enqueue every resource change in the workqueue to trigger a reconcile. This means, the reconcile method gets basically spammed with generic requests without any concrete reference resource to reconcile on. To avoid that you run the expensive logic all the time, you check for the elapsed time and don't trigger when it's not time yet. Have you thought about different ways of solving this issue, like a periodically firing goroutine and populating the controller-runtime informer-cache some other way? The way you implemented it probably works fine, I'm just curious if there isn't a better way.

internal/scheduling/reservations/capacity/controller.go
```diff
@@ -19,52 +20,111 @@ import (
...
-		select {
-		case <-ctx.Done():
-			return nil
-		case <-timer.C:
-			cycleCtx := WithNewGlobalRequestID(ctx)
-			if err := c.reconcileAll(cycleCtx); err != nil {
-				LoggerFromContext(cycleCtx).Error(err, "reconcile cycle failed")
-			}
-			timer.Reset(c.config.ReconcileInterval.Duration)
-		}
```
PhilippMatthes, line 66: What was the idea behind deleting this code and replacing it with a reconcile method? See also my comment https://github.com/cobaltcore-dev/cortex/pull/1025/changes#r3544600386

internal/scheduling/reservations/capacity/controller.go
PhilippMatthes: Doesn't this approach mean that, when the last resource change didn't trigger a reconcileAll because it was just shy of the deadline, but then no further resource change happens, you're getting stale data?

internal/scheduling/reservations/capacity/controller.go
PhilippMatthes, line 559: I noticed the current implementation works a bit differently from a typical reconciler pattern. Traditionally, a reconciler converges a Kubernetes resource's spec (desired state) into its status (actual state), whereas here we have a single routine creating and patching all resources together. Is there a specific reason the CRDs need to be updated in sync? Just trying to understand the context.

I was wondering if it might be worth exploring an approach similar to how we handle cortex datasources? Something like a Start(ctx) routine that discovers all the possible flavor groups and ensures the resource specs exist, paired with a reconciler on each individual flavor group resource that collects the data and writes it to the status. As a nice bonus, we could track a "last updated" timestamp inside the resource and take advantage of the controller-runtime workqueue the way it's designed to be used.

Of course, this is not a show-stopper from my side, but definitely something to be aware of. I can also share more details if you like on how I exactly would implement this, if you want to spend more time on refactoring.

PhilippMatthes (commented):
Amazing job! I would appreciate if we could exchange on my comments a little -- maybe I'm also not grasping the full context. Let me know, we can also sync offline.

internal/scheduling/reservations/capacity/controller.go
juliusclausnitzer: True, more of a defensive programming measure, but not really needed unless someone changes the concurrent reconciles. Since that is not an easily changeable chart value probably overkill

internal/scheduling/reservations/capacity/controller.go
juliusclausnitzer: The work queue should deduplicate natively, so if all resources change simultaneously, it should still only be one call to the reconcile method, which would reduce spam significantly, but of course I understand your point that this seems a bit aggressive. 

Also, the controller-runtime will re-queue after remaining minimum time, so we should not get stale-ness of more than 30 seconds. 

Lets say worst case, 5 resources change in intervals of 5 seconds, and the work queue is emptying so fast that is does not deduplicate here already. In this case the first reconcile goes through. The other 4 would requeue after remaining minimal interval, so all re-queue at the same point in time (after 30s) and would then be deduplicated in the work queue, which would then reconcile and still ensure maximum staleness of 30 sec.

internal/scheduling/reservations/capacity/controller.go
juliusclausnitzer: I think Marcel got annoyed by stale data :) Here is the ticket for a bit more context https://github.wdf.sap.corp/orgs/sap-cloud-infrastructure/projects/36/views/2?sliceBy%5Bvalue%5D=I764822&pane=issue&itemId=236292&issue=sap-cloud-infrastructure%7Cworkload-management-issues%7C490

internal/scheduling/reservations/capacity/controller.go
juliusclausnitzer: I think that might not be optimal here, as we have interdependencies between different flavor groups. For example, to calculate `freeResources` we use a round robin approach that includes all groups. But happy to be proven wrong.

internal/scheduling/reservations/capacity/config.go
```diff
@@ -32,6 +39,9 @@ func (c *Config) ApplyDefaults() {
 	if c.ReconcileInterval.Duration == 0 {
 		c.ReconcileInterval = defaults.ReconcileInterval
 	}
+	if c.MinReconcileInterval.Duration == 0 {
+		c.MinReconcileInterval = defaults.MinReconcileInterval
+	}
```
mblos, line 45: please enforce ReconcileInterval > MinReconcileInterval

internal/scheduling/reservations/capacity/controller.go
```diff
@@ -19,52 +19,106 @@ import (
...
+	}
+
+	c.lastReconcileAt = time.Now()
+
+	return ctrl.Result{RequeueAfter: c.config.ReconcileInterval.Duration}, nil
+}
+
+// SetupWithManager registers the reconciler with the controller manager and sets up watches
+// on all CRDs that affect capacity output. All events are coalesced to a single key.
+func (c *Reconciler) SetupWithManager(mgr ctrl.Manager, mcl *multicluster.Client) error {
```
mblos, line 149: i suggest some small changes to the watch trigger.

- Knowledge: add a predicate to filter "flavor_groups" only
- Hypervisor: add predicates for Allocation, EffectiveCapacity, Capacity, and filter with AZ
- CommittedResource watch: replaced with a Reservation watch. This is more generic and covers all other reservation types as well
- Pipeline watch: kept as-is

internal/scheduling/reservations/capacity/controller.go
```diff
@@ -19,52 +19,106 @@ import (
...
-			timer.Reset(c.config.ReconcileInterval.Duration)
-		}
+// Reconcile implements reconcile.Reconciler. It is called by controller-runtime whenever a
+// watched CRD changes, and also on the periodic RequeueAfter floor set by ReconcileInterval.
+// If called sooner than MinReconcileInterval since the last successful run, it returns early.
+func (c *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
+	elapsed := time.Since(c.lastReconcileAt)
+	minInterval := c.config.MinReconcileInterval.Duration
+
+	if c.lastReconcileAt != (time.Time{}) && elapsed < minInterval {
```
mblos: nit
```suggestion
	if !c.lastReconcileAt.IsZero() && elapsed < minInterval {
```

internal/scheduling/reservations/capacity/config.go
```diff
@@ -9,11 +9,18 @@ import (
...
-	// ReconcileInterval is how often the controller probes the scheduler and updates CRDs.
+	// ReconcileInterval is the periodic floor: how often the reconciler re-runs even without a
+	// watch event. Acts as a fallback for changes not covered by watches (e.g. blocked memory drift).
 	ReconcileInterval metav1.Duration `json:"capacityReconcileInterval"`
 
+	// MinReconcileInterval is the minimum time between two consecutive reconcile runs.
+	// If Reconcile() is called sooner than this since the last successful run, it returns early
+	// with RequeueAfter set to the remaining duration. Prevents back-to-back reconciles on rapid
+	// watch events (e.g. a batch of CommittedResource updates).
+	MinReconcileInterval metav1.Duration `json:"capacityMinReconcileInterval"`
```
mblos, line 23: please add helm values for the config

internal/scheduling/reservations/capacity/controller.go
```diff
@@ -19,52 +19,106 @@ import (
...
+
+	c.lastReconcileAt = time.Now()
+
+	return ctrl.Result{RequeueAfter: c.config.ReconcileInterval.Duration}, nil
+}
+
+// SetupWithManager registers the reconciler with the controller manager and sets up watches
+// on all CRDs that affect capacity output. All events are coalesced to a single key.
+func (c *Reconciler) SetupWithManager(mgr ctrl.Manager, mcl *multicluster.Client) error {
+	log.Info("starting capacity reconciler",
```
mblos, line 150: please add new config values here

mblos (changes requested):
overall looks good, thanks. Mainly some small changes for the watch triggers suggested

internal/scheduling/reservations/capacity/controller.go
juliusclausnitzer: which new config values?

mblos: capacityMinReconcileInterval

mblos (dismissed):

internal/scheduling/reservations/capacity/controller.go
```diff
@@ -19,52 +19,160 @@ import (
...
+		return !capacityMapsEqual(oldHV.Status.Allocation, newHV.Status.Allocation) ||
+			!capacityMapsEqual(oldHV.Status.EffectiveCapacity, newHV.Status.EffectiveCapacity) ||
+			!capacityMapsEqual(oldHV.Status.Capacity, newHV.Status.Capacity)
+	},
+	DeleteFunc: func(e event.DeleteEvent) bool {
+		_, hasAZ := e.Object.GetLabels()["topology.kubernetes.io/zone"]
+		return hasAZ
+	},
+	GenericFunc: func(e event.GenericEvent) bool { return false },
+}
```
mblos, line 84: Nit, we could do more fine granular checks, but fine as it is


```suggestion
var hvCapacityChangePredicate = predicate.Funcs{
	capacityEqual := func(a, b map[hv1.ResourceName]resource.Quantity) bool {
		if len(a) != len(b) {
			return false
		}
		for k, av := range a {
			bv, ok := b[k]
			if !ok || av.Cmp(bv) != 0 {
				return false
			}
		}
		return true
	}
	changed := func(oldObj, newObj client.Object) bool {
		o, ok1 := oldObj.(*hv1.Hypervisor)
		n, ok2 := newObj.(*hv1.Hypervisor)
		if !ok1 || !ok2 {
			return true
		}
		if o.Labels["topology.kubernetes.io/zone"] != n.Labels["topology.kubernetes.io/zone"] {
			return true
		}
		if !capacityEqual(o.Status.Capacity, n.Status.Capacity) {
			return true
		}
		if !capacityEqual(o.Status.EffectiveCapacity, n.Status.EffectiveCapacity) {
			return true
		}
		if !capacityEqual(o.Status.Allocation, n.Status.Allocation) {
			return true
		}
		return false
	}
	return predicate.Funcs{
		CreateFunc:  func(_ event.CreateEvent) bool { return true },
		UpdateFunc:  func(e event.UpdateEvent) bool { return changed(e.ObjectOld, e.ObjectNew) },
		DeleteFunc:  func(_ event.DeleteEvent) bool { return true },
		GenericFunc: func(_ event.GenericEvent) bool { return false },
	}
}
```

mblos (dismissed):

mblos (approved):

PR #1042, opened 2026-07-14
feat: pending cache for configurable list of gvks

SoWieMarkus:
Adds a transparent, in-process write-through cache that masks informer lag: after a write the controller can immediately Get/List the object even before the informer has caught up.

So what this cache does is:
- On Create/Update/Patch stores an entry in the cache (if the gvk is configured); on Delete sets a tombstone.
- Reads merge the overlay with the informer result: a tombstone yields NotFound, a live entry wins over (or supplements) the informer, and List/MatchingFields include pending entries.
- Entries are evicted once the real object shows up in the informer at a resourceVersion ≥ the cached one, or after a TTL as a safety net.

This is integrated into the multicluster client:
- multicluster.Client wraps each home/remote cluster.Cluster via pendingcache.WrapCluster, so every cluster gets its own overlay + informer and eviction stays cluster-local. Routing is unchanged - reads/writes/IndexField just flow through the wrapped cluster's GetClient()/GetFieldIndexer().
- The overlay's lifecycle (eviction handlers + TTL cleanup) is a manager.Runnable (NeedLeaderElection = false) added alongside the raw cluster, so the informer is started exactly once.

The gvks that are supposed to be cached and other config options are listed below:
```yml
cache:
  enabled: false                              # opt-in; off ⇒ raw clusters, zero overhead
  gvks: [cortex.cloud/v1alpha1/Reservation]   # only these GVKs are overlaid; others pass through
  ttl: 2m                                      # max entry lifetime (default 2m)
```

```mermaid
  flowchart TB
      Ctrl["Controller"]
      MCL["multicluster.Client\n(routing / fan-out)"]

      subgraph home["overlayCluster (home)"]
          OH["Overlay (cache)"]
          CH["cluster + informer"]
          OH --> CH
      end
      subgraph rem["overlayCluster (remote N)"]
          OR["Overlay (cache)"]
          CR["cluster + informer"]
          OR --> CR
      end

      Ctrl --> MCL
      MCL --> OH
      MCL --> OR

      CH -. "informer events → evict" .-> OH
      CR -. "informer events → evict" .-> OR
```


pkg/clientcache/cache.go
```diff
@@ -0,0 +1,262 @@
...
+		return nil, false
+	}
+	e, ok := entries[key]
+	return e, ok
+}
+
+// overlayList merges the overlay entries for the GVK into the informer result,
+// deduplicating by objectKey (overlay wins), dropping tombstones, and filtering
+// overlay-only entries against the list options' label and field selectors.
+func (o *overlay) overlayList(gvk schema.GroupVersionKind, existing []runtime.Object, lo *client.ListOptions) []runtime.Object {
```
mblos: If a write changes an indexed field, the inner List filters against the informer (old value), then overlayList silently replaces the object with the new version.

Example: informer has spec.hostName=host-A (RV=5), overlay has spec.hostName=host-B (RV=6). List(MatchingFields{"spec.hostName": "host-A"}) includes the object from the informer, then overlayList swaps in the overlay version.

So you get a host-B object in a host-A query.
Or am I overlooking something?

pkg/clientcache/runnable.go
```diff
@@ -0,0 +1,84 @@
...
+			return nil
+		case now := <-ticker.C:
+			c.overlay.cleanupExpired(now)
+		}
+	}
+}
+
+// evictionHandler returns an informer event handler that evicts overlay entries
+// for the GVK when the real object is observed at a >= ResourceVersion.
+func (c *CachingClient) evictionHandler(gvk schema.GroupVersionKind) toolscachek8s.ResourceEventHandler {
```
mblos, line 67: is DeleteFunc missing here? what if eg someone via kubectl deletes the crd, the informer fires a delete .. but the overlay does not react on it, or?

pkg/clientcache/client.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	if gvk, cached := c.gvkFor(obj); cached {
+		c.overlay.upsert(gvk, obj)
+	}
+	return nil
+}
+
+// Delete delegates to the inner client and, on success for a cached GVK, stores
+// a tombstone in the overlay.
+func (c *CachingClient) Delete(ctx context.Context, obj client.Object, opts ...client.DeleteOption) error {
```
mblos, line 494: what about DeleteAllOf

pkg/clientcache/cache.go
```diff
@@ -0,0 +1,262 @@
...
+func newOverlay(ttl time.Duration) *overlay {
+	return &overlay{
+		byGVK:    make(map[schema.GroupVersionKind]map[objectKey]*entry),
+		ttl:      ttl,
+		indexers: make(map[schema.GroupVersionKind]map[string]client.IndexerFunc),
+	}
+}
+
+func keyForObject(obj client.Object) objectKey {
+	return objectKey{namespace: obj.GetNamespace(), name: obj.GetName()}
```
mblos: What if a crd resource has the same name across lh clusters?

mblos (commented):
Very nice work :) some first comments (still not completed with reviewing)

cmd/manager/main.go
```diff
@@ -519,8 +536,8 @@ func main() {
 	}
 	if slices.Contains(mainConfig.EnabledControllers, "hypervisor-overcommit-controller") {
 		hypervisorOvercommitController := &nova.HypervisorOvercommitController{}
-		hypervisorOvercommitController.Client = multiclusterClient
-		if err := hypervisorOvercommitController.SetupWithManager(mgr); err != nil {
+		hypervisorOvercommitController.Client = cachingClient
+		if err := hypervisorOvercommitController.SetupWithManager(mgr, multiclusterClient); err != nil {
```
PhilippMatthes, line 556: Can this cause issues? Inside the SetupWithManager function, aren't there calls that would need to be tunneled by the caching client, such as defining indexes or resource handlers?
Is it possible to wrap it the other way around? Instead of ctrl.Client -> mcl.Client -> clientcache.Client, do ctrl.Client -> clientcache.Client -> mcl.Client? This wouldn't break the pattern here

helm/bundles/cortex-nova/values.yaml
```diff
@@ -110,6 +110,9 @@ cortex: &cortex
           - kvm.cloud.sap/v1/Hypervisor
           - kvm.cloud.sap/v1/HypervisorList
           - v1/Secret
+    clientcache:
+      gvks:
+        - cortex.cloud/v1alpha1/Reservation
     keystoneSecretRef:
```
PhilippMatthes, line 116: Is it expensive to cache resources? If not, why not cache all resources w/o extra configuration?

pkg/clientcache/client.go
```diff
@@ -0,0 +1,265 @@
...
+			return nil, errors.New("clientcache: no gvk registered in scheme for " + s)
+		}
+		out[gvk] = true
+	}
+	return out, nil
+}
+
+// Inner returns the wrapped client, e.g. for use with a controller Builder that
+// needs the raw client rather than the caching wrapper.
+func (c *CachingClient) Inner() client.Client { return c.Client }
```
PhilippMatthes: CachingClient is not an interface. What purpose does this method have?

pkg/clientcache/client.go
```diff
@@ -0,0 +1,265 @@
...
+// refreshes the overlay entry.
+func (c *CachingClient) Update(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
+	if err := c.Client.Update(ctx, obj, opts...); err != nil {
+		return err
+	}
+	if gvk, cached := c.gvkFor(obj); cached {
+		c.overlay.upsert(gvk, obj)
+	}
+	return nil
+}
```
PhilippMatthes, line 470: What happens when there are two concurrent calls to this function?
Can there be a race condition w.r.t the client update executing in another order than the cache update? Ideally, we should add a globally shared mutex across all CRUD functions so that concurrently reading/writing routines see a consistent state.

pkg/clientcache/cache.go
```diff
@@ -0,0 +1,262 @@
...
+// objectKey identifies an object within a GVK by namespace and name.
+type objectKey struct {
+	namespace string
+	name      string
+}
+
+// overlay is the generic, client-independent core of the cache. It stores
+// pending writes keyed by GVK and objectKey and merges them into informer
+// read results until the real object is observed in an informer (eviction).
+type overlay struct {
```
PhilippMatthes:
```suggestion
type cache struct {
```

pkg/clientcache/client.go
```diff
@@ -0,0 +1,265 @@
...
+
+// Patch delegates to the inner client and, on success for a cached GVK,
+// refreshes the overlay entry with the patched object.
+func (c *CachingClient) Patch(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
+	if err := c.Client.Patch(ctx, obj, patch, opts...); err != nil {
+		return err
+	}
+	if gvk, cached := c.gvkFor(obj); cached {
+		c.overlay.upsert(gvk, obj)
```
PhilippMatthes: Is it intended that this method call cannot fail when the expected resource has not been created yet? Doesn't this violate the expected controller-runtime client interface protocol?

pkg/clientcache/client.go
```diff
@@ -0,0 +1,265 @@
...
+// the inner client unchanged.
+type CachingClient struct {
+	client.Client // inner client, used for delegation
+
+	informers InformerSource
+	scheme    *runtime.Scheme
+	overlay   *overlay
+	ttl       time.Duration
+	gvks      map[schema.GroupVersionKind]bool
+}
```
PhilippMatthes, line 168: I would prefer if we flatten out the caching logic into the client, and not have a separate cache struct.
This would make it easier to understand where the resources stored in the map locally deviate from the resources stored in the controller-runtime client.
For example, the Delete method won't delete the resource from the map immediately -- it will set a deleted flag on the resource. What happens from there, no clue.

pkg/clientcache/client.go
```diff
@@ -0,0 +1,265 @@
...
+
+// New builds a CachingClient wrapping inner. informers supplies the informers
+// used for eviction, scheme resolves object GVKs, and conf lists the GVKs to
+// overlay and the TTL. GVK strings are formatted as "<group>/<version>/<Kind>"
+// and are resolved against scheme.
+func New(inner client.Client, informers InformerSource, scheme *runtime.Scheme, conf Config) (*CachingClient, error) {
```
PhilippMatthes: If I see it correctly, you are inputting New(mcl.Client, mcl.Client) into this function in main.go -- what's the purpose of mapping out the InformerSource as an additional interface instead of just taking client.Client?

pkg/clientcache/runnable.go
```diff
@@ -0,0 +1,84 @@
...
+	ro, err := c.scheme.New(gvk)
+	if err != nil {
+		return nil, err
+	}
+	obj, ok := ro.(client.Object)
+	if !ok {
+		return nil, fmt.Errorf("clientcache: object for gvk %s does not implement client.Object", gvk)
+	}
+	return obj, nil
+}
```
PhilippMatthes, line 99: Any specific reason why this code is outside client.go?

PhilippMatthes (changes requested):
Thank you for incorporating my feedback from https://github.com/cobaltcore-dev/cortex/pull/1015 -- especially, the informer-cache idea and reusing metav1.Duration! I've copied over some thoughts and questions and had some new ones along the way. Thanks for considering my feedback.

cmd/manager/main.go
SoWieMarkus: I did it that way, because multiclusterclient has multiple clients. All of these would need to be wrapped.

But I see the issue with the index.

helm/bundles/cortex-nova/values.yaml
SoWieMarkus: For now just to test how expensive it really is. We can change that afterwards

pkg/clientcache/client.go
SoWieMarkus: Dead code that I forgot to remove. Change is applied locally

pkg/clientcache/client.go
SoWieMarkus: updated the interface locally. main purpose is to avoid importing mcl.

pkg/clientcache/client.go
SoWieMarkus: Can you maybe explain a little bit more what you concern is here, because I think I don't understand?
In Patch and Update we first call the inner client method. If that fails, we early return with the error. If there is no error we apply the change also in our cache.

pkg/clientcache/runnable.go
SoWieMarkus: The split keeps data structure logic in client.go and controller-runtime lifecycle (manager.Runnable, informer wiring, ticker) in a separate file. I think that makes the code easier to understand. But no strong opinion on that. Can also move this to client.go

pkg/clientcache/client.go
SoWieMarkus: Added a mutex

pkg/clientcache/cache.go
SoWieMarkus: Fair point. Do you have any idea how to resolve that?

I mean we would have to extend objectKey with a cluster discriminator. Problem: the cache is deliberately decoupled from multicluster and Get/List don't carry any ref to the cluster in their requests. Writes also only map that via the routers of the multi cluster client. 

Only "solution" I see is to merge cache and multi cluster client, but I don't think that we want that.

pkg/clientcache/client.go
SoWieMarkus: I mean I can add that. Just thought we don't need it.

pkg/clientcache/client.go
PhilippMatthes: If the resource is not inside the cache, this points toward a data inconsistency and the method should fail instead. Not?

SoWieMarkus: The cache only tracks objects that were written during runtime. So if I understand your point correctly, that is the correct behavior

pkg/clientcache/cache.go
SoWieMarkus: as discussed in the weekly we treat the mcl as logically one cluster which implements these rules. this resolves this issue

pkg/multicluster/client.go
```diff
@@ -132,13 +99,34 @@ func (c *Client) InitFromConf(ctx context.Context, mgr ctrl.Manager, conf Client
...
+	// Runnable and must NOT re-Start the inner home cluster.
+	if c.cacheConf.Enabled && c.HomeCluster != nil {
+		wrapped, overlay, err := pendingcache.WrapCluster(c.HomeCluster, c.cacheConf)
+		if err != nil {
+			return err
+		}
+		c.HomeCluster = wrapped
+		if err := mgr.Add(overlay); err != nil {
+			return err
+		}
```
PhilippMatthes: I would pull this out of the multicluster client and make it generic, e.g. like this:
```go
type ClusterWrapper interface{
  WrapCluster(cl cluster.Cluster) cluster.Cluster
}
```
Then add this field to the multicluster client:
```go
type Client struct {
  // ...
  Wrappers []ClusterWrapper
}
```
and then just loop over this slice where ever you like to mutate the cluster.

In our case, you would then implement a cluster wrapper for your cache inside the cache package, and add this to the multicluster client inside the main function.

PhilippMatthes: Other idea:
```go
func (c *Client) MutateClusters(f func ([]cluster.Cluster) cluster.Cluster) {
  // Apply f to every cluster...
}
```

helm/bundles/cortex-nova/values.yaml
```diff
@@ -110,6 +110,10 @@ cortex: &cortex
           - kvm.cloud.sap/v1/Hypervisor
           - kvm.cloud.sap/v1/HypervisorList
           - v1/Secret
+    pendingcache:
```
PhilippMatthes: Why `pendingcache` and not just `cache`? What's the meaning of `pending`?

pkg/multicluster/client.go
```diff
@@ -53,55 +56,19 @@ type Client struct {
...
-	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
-	// The resource GVKs this apiserver serves, formatted as "<group>/<version>/<Kind>".
-	GVKs []string `json:"gvks"`
-	// Labels used by ResourceRouters to match resources to this cluster
-	// for write operations (Create/Update/Delete/Patch).
-	Labels map[string]string `json:"labels,omitempty"`
+	// cacheConf configures the optional in-process overlay cache. When
+	// Enabled, each home/remote cluster is wrapped so its GetClient/
+	// GetFieldIndexer route through a per-cluster overlay. Set in InitFromConf.
+	cacheConf pendingcache.Config
```
PhilippMatthes: See https://github.com/cobaltcore-dev/cortex/pull/1042/changes#r3812678525, can be pulled out

pkg/multicluster/client.go
```diff
@@ -181,11 +188,13 @@ func (c *Client) AddRemote(ctx context.Context, host, caCert string, insecureSki
...
 		c.remoteClusters[gvk] = append(c.remoteClusters[gvk], remoteCluster{
-			cluster: cl,
+			cluster: stored,
 			labels:  labels,
 		})
 	}
-	return cl, nil
+	// Return the raw inner cluster so the caller starts its informers/caches;
+	// the overlay Runnable (if any) is returned separately.
+	return cl, overlay, nil
```
PhilippMatthes: I would prefer if we can keep the multicluster client largely untouched by the cache changes.

pkg/pendingcache/client.go
```diff
@@ -0,0 +1,662 @@
...
+	key objectKey
+}
+
+// refMutex is the actual per-object lock plus a reference count of how many
+// callers currently hold or are waiting for it. The count lets keyedMutex know
+// when the entry is unused so it can be deleted (see keyedMutex.lock).
+type refMutex struct {
+	mu  sync.Mutex
+	ref int
+}
```
PhilippMatthes: Isn't this the same as https://pkg.go.dev/golang.org/x/sync/semaphore ?

pkg/pendingcache/client.go
```diff
@@ -0,0 +1,662 @@
...
+
+// objectKey identifies an object within a GVK by namespace and name.
+type objectKey struct {
+	namespace string
+	name      string
+}
+
+func keyForObject(obj client.Object) objectKey {
+	return objectKey{namespace: obj.GetNamespace(), name: obj.GetName()}
+}
```
PhilippMatthes: Why not use `types.NamespacedName` from `k8s.io/apimachinery/pkg/types`?

PhilippMatthes: There's also `client.ObjectKey = types.NamespacedName` from `"sigs.k8s.io/controller-runtime/pkg/client"` which implements a method `client.ObjectKeyFromObject` you could reuse.

pkg/pendingcache/client.go
```diff
@@ -0,0 +1,662 @@
...
+//   - k.mu guards the locks map itself. It is only ever held for the tiny
+//     bookkeeping critical sections below (map lookup/insert/delete and the
+//     ref counter), never across the caller's I/O.
+//   - rm.mu is the real per-object lock, held by the caller across the inner
+//     client call and the overlay mutation.
+//
+// Because k.mu is never held while rm.mu is locked (we release k.mu before
+// taking rm.mu), the two can never deadlock against each other.
+type keyedMutex struct {
+	mu    sync.Mutex
```
PhilippMatthes: locksMu?

pkg/cache/runnable.go
```diff
@@ -0,0 +1,96 @@
...
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case now := <-ticker.C:
+			c.cleanupExpired(now)
+		}
+	}
+}
```
PhilippMatthes, line 52: Hmm, do we really need to poll here? Why not spawn a go routine when the entry is created, which sleeps for the precise amount of time needed (+ some random jitter), until it is executed.

pkg/pendingcache/client.go
SoWieMarkus: Thanks for that

pkg/cache/runnable.go
SoWieMarkus: Tried this approach initially but discarded it since the informer evicts first in the common case, the per-entry goroutine becomes pure overhead.
It needs to be cancelled on eviction (and on updates, the old timer must be stopped before starting a new one), adding a lot of complexity for a path that rarely fires. A periodic sweep - imo - is much simpler and the imprecision doesn't matter for a safety net.

pkg/pendingcache/client.go
SoWieMarkus: The semaphore counter answers "how many more callers may enter right now." Once someone acquires it, the count drops but it knows nothing about the goroutines still waiting in line behind it.

Here the ref counter answers a different question: "is anyone still interested in this entry at all?" It counts both the current holder and every goroutine blocked waiting for it. Only when ref hits zero can you safely delete the map entry, because you know nobody is waiting either.

Or am I missing something?

helm/bundles/cortex-nova/values.yaml
SoWieMarkus: Pending = Waiting for informer
Thats how I called the cache since just "cache" could mean everything and might be confused with the kubernetes one.
But for simplicity we can rename that.

PhilippMatthes: How about writecache? That's what it does in essence no? Cache local writes?

PhilippMatthes: We could also name the package cache and the struct WriteCache or something like this

pkg/cache/runnable.go
PhilippMatthes: Makes sense, thanks for explaining

pkg/multicluster/client.go
```diff
@@ -20,8 +20,22 @@ import (
...
+// cluster before it is stored for routing. WrapCluster receives the manager and
+// the current cluster: raw for the first wrapper in the chain, or the previous
+// wrapper's result for subsequent ones. It must return the (possibly wrapped)
+// cluster to use for routing and is responsible for registering any lifecycle
+// Runnables with mgr itself. Wrappers are applied in order for every home and
+// remote cluster during InitFromConf. For remote clusters the original unwrapped
+// cluster is always added to the manager separately so its informers start
+// independently of wrapping.
+type ClusterWrapper interface {
+	WrapCluster(mgr manager.Manager, cl cluster.Cluster) (cluster.Cluster, error)
```
PhilippMatthes: Can we not remove `mgr` from the `ClusterWrapper` interface? This leads to the `AddRemote` method needs changing, which I would like to keep the same so other consumers of our package don't break. Can't we pass the manager to the cache client? Otherwise, we still can wrap the clusters with a closure function (the 2nd approach I proposed), which can capture the manager as outside context.

PhilippMatthes (approved):

PR #1048, opened 2026-07-15
feat(reservations): follow VM live migration in CR reservation reconciler

juliusclausnitzer:
## Summary

Implements issue #373. When the reservation reconciler detects that a
confirmed VM has disappeared from its expected host, it now searches all
hypervisor CRDs to determine whether the VM live-migrated rather than
treating it as terminated.

- **New host has capacity**: `Spec.TargetHost` is updated to the new
  host. The existing TargetHost→Status.Host sync path (Branch B in
  `Reconcile`) advances `Status.Host` on the next cycle, keeping
  capacity accounting consistent during the transition.
- **New host has no capacity**: `TargetHost` is left unchanged; the
  VM's actual location is recorded in `Status.Allocations`; the new
  `VMMisplaced` condition is set with reason `MigratedToFullHost`.
- **VM not found anywhere**: existing stale-removal behaviour,
  unchanged.

Only the migrating VM's host is considered per-reconcile. Other
allocated VMs in the same reservation are not moved.

The HV list is fetched lazily — only when a confirmed VM is actually
missing from its expected host — so the common path (no migration)
pays no extra API cost.


mblos (changes requested):
Thanks for the PR. Some things have changed since ticket was updated, so I suggest a slightly refined (and simplified) semantic as follows, wdyt: 
- Single VM in reservation: if new host fits -> follow (update TargetHost). If not -> remove VM from reservation (slot stays on old host, available for re-use).
- Multiple VMs in reservation: always remove the migrated VM, never update TargetHost. moving it would make the remaining VMs appear "misplaced" on the next reconcile.

Some consequences:
1. This makes the `VMMisplaced ` condition obsolete, we can remove it. Good to have a proper log message to note what has happened when CR VMs are migrated. For the cases where we used the misplaced condition, we can simply remove the VM from the reservation.
2. We can guard the TargetHost update with only for `len(allocated VMs) == 1`.

More complex cases like cross-reservation reassignment, re-mapping of VMs, .. are considered when the VM crd exists.

internal/scheduling/reservations/capacity_accounting.go
```diff
@@ -10,6 +10,71 @@ import (
...
+		other := &allReservations[i]
+		if other.Name == res.Name && other.Namespace == res.Namespace {
+			continue
+		}
+		// Only block resources from reservations that target or are confirmed on this host.
+		targetsThisHost := other.Spec.TargetHost == hv.Name || other.Status.Host == hv.Name
+		if !targetsThisHost {
+			continue
+		}
+		for rn, block := range UnusedReservationCapacity(other, false) {
```
mblos: or some other readable name
```suggestion
		for resourceName, block := range UnusedReservationCapacity(other, false) {
```

mblos (dismissed):

mblos (approved):

PR #1054, opened 2026-07-16
feat: add kpi that tracks CR count per configured cluster

SoWieMarkus:
- Add `CountPerGVK` method to multicluster client
- This uses the PartialObjectMetadataList so only object metadata is fetched, making it efficient even for large object counts
- Counts the number of objects per cluster and exports it as metric with the configured labels of the remote cluster


pkg/multicluster/client.go
```diff
@@ -183,6 +184,32 @@ func (c *Client) AddRemote(ctx context.Context, host, caCert string, insecureSki
 	return cl, nil
 }
 
+// AddRemoteCluster registers a pre-built cluster.Cluster as a remote for the
+// given GVKs with the provided routing labels. Unlike AddRemote, it does not
+// create a new cluster from a REST config — the caller supplies the cluster
+// directly. This is useful for in-process wiring and for unit tests.
+func (c *Client) AddRemoteCluster(cl cluster.Cluster, labels map[string]string, gvks ...schema.GroupVersionKind) {
```
PhilippMatthes: I cannot find any references to this function (apart from the unit test) -- stale code?

pkg/multicluster/client.go
```diff
@@ -183,6 +184,32 @@ func (c *Client) AddRemote(ctx context.Context, host, caCert string, insecureSki
...
+	for _, gvk := range gvks {
+		c.remoteClusters[gvk] = append(c.remoteClusters[gvk], remoteCluster{
+			cluster: cl,
+			labels:  labels,
+		})
+	}
+}
+
+// AddHomeGVK marks the given GVK as served by the home cluster.
+func (c *Client) AddHomeGVK(gvk schema.GroupVersionKind) {
```
PhilippMatthes: Same for this code: is this only used by the unit test? Then, we should put it in the unit test.

pkg/multicluster/client.go
```diff
@@ -448,6 +475,62 @@ func (c *Client) Apply(ctx context.Context, obj runtime.ApplyConfiguration, opts
...
+	Count  int
+	IsHome bool
+}
+
+// CountPerClusterByGVK returns the number of objects of the given GVK in each
+// configured cluster. It uses PartialObjectMetadataList so only object metadata
+// crosses the wire — no spec or status — making it efficient even for large
+// object counts. Clusters that return an error are logged and skipped (same
+// policy as List). The home cluster is included with IsHome set to true.
+func (c *Client) CountPerClusterByGVK(ctx context.Context, gvk schema.GroupVersionKind, opts ...client.ListOption) ([]ClusterObjectCount, error) {
```
PhilippMatthes: Should we make this a little more generic? Something like: `ListByCluster` and then put the counting logic in the kpi?

PhilippMatthes (commented):
Looks good! Just some minor comments.

PhilippMatthes (dismissed):
zack peng peng

PhilippMatthes (approved):

PR #1067, opened 2026-07-20
Filter kvm hypervisors for non-kvm images

PhilippMatthes:
This adds a new Nova scheduler filter step, filter_image_properties, which removes known KVM hypervisors from the set of scheduling candidates when the image being booted declares a non-KVM hypervisor type. Some images are pinned to a specific hypervisor such as VMware or baremetal, and scheduling them onto KVM hosts would fail late in the boot process. Filtering these hosts out early during scheduling avoids wasted placement attempts and produces better placement decisions.

To support this, the Nova external API gains a NovaImageMetaHVType type together with constants for the vmware, baremetal, and kvm hypervisor types, and a NovaImageMeta.GetHypervisorType helper that reads the img_hv_type and hypervisor_type image properties and normalizes them into that type. The new FilterImagePropertiesStep uses this helper to decide, per request, whether the image requires a hypervisor type other than KVM, and if so it lists the available hypervisors and drops the known KVM hosts from the activations.

The step is registered in the KVM scheduling pipelines via helm/bundles/cortex-nova/templates/pipelines_kvm.yaml so it takes effect for KVM deployments. Test coverage includes unit tests for the new GetHypervisorType helper and a dedicated test suite for the filter step covering the filtering behavior.

Additionally, the monitoring library for steps is extended such that events can be surfaced as a prometheus metric -- and we now use this to alert if too many images are tried without a hypervisor type in their properties.

Assisted-by: Claude Code:claude-opus-4-1-20250805 [Bash] [Read]

### DIFF
--- a/api/external/nova/messages.go
+++ b/api/external/nova/messages.go
@@ -315,6 +315,46 @@ type NovaImageMeta struct {
 	Properties      NovaObject[map[string]any] `json:"properties"`
 }
 
+type NovaImageMetaHVType string
+
+const (
+	NovaImageMetaHVTypeVMware    NovaImageMetaHVType = "vmware"
+	NovaImageMetaHVTypeBaremetal NovaImageMetaHVType = "baremetal"
+	NovaImageMetaHVTypeKVM       NovaImageMetaHVType = "kvm"
+)
+
+// GetHypervisorType determines the hypervisor type based on the image properties.
+func (m NovaImageMeta) GetHypervisorType() (NovaImageMetaHVType, error) {
+	if m.Properties.Data == nil {
+		return "", errors.New("image properties are not set")
+	}
+	keys := []string{
+		"img_hv_type",     // Used by kvm images
+		"hypervisor_type", // Used by other images (e.g., VMware)
+	}
+	for _, key := range keys {
+		val, ok := m.Properties.Data[key]
+		if !ok {
+			continue
+		}
+		str, ok := val.(string)
+		if !ok {
+			return "", fmt.Errorf("hypervisor type in image properties is not a string: %v", val)
+		}
+		switch strings.ToLower(str) {
+		case "vmware":
+			return NovaImageMetaHVTypeVMware, nil
+		case "baremetal":
+			return NovaImageMetaHVTypeBaremetal, nil
+		case "kvm":
+			return NovaImageMetaHVTypeKVM, nil
+		default:
+			return "", fmt.Errorf("unsupported hypervisor type in image properties: %s", str)
+		}
+	}
+	return "", errors.New("hypervisor type not found in image properties")
+}
+
 // Nova flavor metadata for the specified VM.
 type NovaFlavor struct {
 	ID          int               `json:"id"`
--- a/api/external/nova/messages.go
+++ b/api/external/nova/messages.go
@@ -329,7 +329,9 @@ func (m NovaImageMeta) GetHypervisorType() (NovaImageMetaHVType, error) {
 		return "", errors.New("image properties are not set")
 	}
 	keys := []string{
-		"img_hv_type",     // Used by kvm images
+		// See https://github.com/sapcc/nova/blob/27eb1f0/nova/scheduler/filters/image_props_filter.py#L55
+		"img_hv_type", // Used by kvm images
+		// See https://github.com/sapcc/nova/blob/27eb1f0/doc/source/admin/configuration/hypervisor-vmware.rst#tag-vmware-images
 		"hypervisor_type", // Used by other images (e.g., VMware)
 	}
 	for _, key := range keys {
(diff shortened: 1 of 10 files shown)


internal/scheduling/nova/plugins/filters/filter_image_properties.go
```diff
@@ -0,0 +1,50 @@
...
+			"image_hypervisor_type", hvType)
+		hvs := new(hv1.HypervisorList)
+		if err := s.Client.List(context.Background(), hvs); err != nil {
+			traceLog.Error("failed to list hypervisors", "error", err)
+			return nil, err
+		}
+		for _, hv := range hvs.Items {
+			delete(result.Activations, hv.Name)
+			traceLog.Info("filtering host which is kvm hypervisor", "host", hv.Name)
+		}
```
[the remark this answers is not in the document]
PhilippMatthes: Good suggestion, I'll reduce the log spam

PhilippMatthes:
Don't know why the pull-request-creator agent checked out a claude/ branch -- the logic is implemented by me :D

api/external/nova/messages.go
```diff
@@ -315,6 +315,46 @@ type NovaImageMeta struct {
...
+	NovaImageMetaHVTypeBaremetal NovaImageMetaHVType = "baremetal"
+	NovaImageMetaHVTypeKVM       NovaImageMetaHVType = "kvm"
+)
+
+// GetHypervisorType determines the hypervisor type based on the image properties.
+func (m NovaImageMeta) GetHypervisorType() (NovaImageMetaHVType, error) {
+	if m.Properties.Data == nil {
+		return "", errors.New("image properties are not set")
+	}
+	keys := []string{
```
umswmayj, line 331: Optional: It would be create to have a link to where those 2 keys come from.

helm/bundles/cortex-nova/templates/pipelines_kvm.yaml
```diff
@@ -781,6 +796,11 @@ spec:
         This step will consider the `ignore_hosts` and `force_hosts` instructions
         from the nova scheduler request spec to filter out or exclusively allow
         certain hosts.
+    - name: filter_image_properties
```
umswmayj, line 799: afaik (also did a quick grep) the `*-no-history` pipelines are not used anymore (also `*failover-reservation` are not used). 

Not sure if we want to remove them in this pr but we should probably not update the unused pipelines

internal/scheduling/nova/plugins/filters/filter_image_properties.go
```diff
@@ -0,0 +1,50 @@
...
+// Filter hosts based on image properties given in the request spec.
+func (s *FilterImagePropertiesStep) Run(traceLog *slog.Logger, request api.ExternalSchedulerRequest) (*lib.FilterWeigherPipelineStepResult, error) {
+	result := s.IncludeAllHostsFromRequest(request)
+	// If the image properties indicate any other hypervisor type than kvm,
+	// we filter out all known kvm hypervisors.
+	hvType, err := request.Spec.Data.Image.Data.GetHypervisorType()
+	if err != nil {
+		// Hypervisor type could not be determined. In this case, just
+		// ignore the filter and return all hosts.
+		traceLog.Warn("could not determine hypervisor type from image properties",
```
umswmayj, line 28: A metric would be good

umswmayj (dismissed):

helm/bundles/cortex-nova/templates/pipelines_kvm.yaml
PhilippMatthes: Let's keep this out of this PR.

api/external/nova/messages.go
PhilippMatthes: Added in https://github.com/cobaltcore-dev/cortex/pull/1067/commits/ccdce3f8673ef40c49143ba572c7ebd7cee8019e

internal/scheduling/nova/plugins/filters/filter_image_properties.go
PhilippMatthes: This'll require some extension of the library code, but why not do this now instead of later.

PhilippMatthes: Implemented in https://github.com/cobaltcore-dev/cortex/pull/1067/commits/af8624e20e3c3048142717b4a6fc0bab0dfbb63d

PhilippMatthes:
@coderabbit please green check mark

internal/scheduling/nova/plugins/filters/filter_image_properties.go
umswmayj: I like it and it is probably better then my workaround in filter_quota_enforcement_metrics.
In a followup we may can make it a bit more generic. Just noticed that `WithLabelValues` can panic maybe something we need to check if that can happen for us

umswmayj (approved):

PR #1097, opened 2026-07-28
Guard multicluster Create against cross-cluster name collisions

PhilippMatthes:
Writes routed to a single cluster via clusterForWrite did not check whether the resulting object name would collide with an existing object on another cluster serving the same GVK. This produced errors, observed in practice for failover reservations. This change makes Create do a best-effort cross-cluster existence check before writing and return a duplicate error (checkable via IsDuplicateError) instead of creating a colliding object. A Prometheus counter cortex_multicluster_cross_cluster_name_conflicts_total, labeled by method and gvk, is incremented from all duplicate-detection paths. The nil-safe Monitor is wired into the nova multicluster client and registered with the metrics registry, and a CortexNovaMulticlusterNameConflicts alert fires on the counter. The placement shim now also constructs and attaches the monitor to its multicluster client in the production setup path, registering it with the wrapped metrics registry, and ships a matching CortexPlacementShimMulticlusterNameConflicts alert scoped to its metrics service. The e2e path uses a metrics-disabled manager and is unchanged.

Assisted-by: Claude Code:claude-opus-latest [Bash] [Read]


helm/bundles/cortex-nova/templates/alerts.yaml
```diff
@@ -669,4 +669,26 @@ spec:
           The committed resource quota API (Limes LIQUID integration) is returning
           HTTP 5xx errors. This indicates internal problems computing or applying
           quota. Limes may not be able to enforce committed resource quotas.
+
+    - alert: CortexNovaMulticlusterNameConflicts
+      expr: |
+        sum by (method, gvk) (increase(cortex_multicluster_cross_cluster_name_conflicts_total{service="cortex-nova-metrics"}[5m])) > 0
+      for: 5m
```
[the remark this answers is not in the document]
PhilippMatthes: Good catch — fixed by widening the `increase()` lookback from `[5m]` to `[10m]` so the window comfortably exceeds the `for: 5m` pending period, letting a single recorded conflict fire reliably. Applied to both the nova alert and the equivalent `CortexPlacementShimMulticlusterNameConflicts` alert in the placement-shim bundle.

umswmayj (dismissed):

pkg/multicluster/monitor.go
```diff
@@ -0,0 +1,52 @@
...
+	// detected on more than one cluster serving the GVK, labeled by the method
+	// of access and the resource GVK.
+	crossClusterNameConflicts *prometheus.CounterVec
+}
+
+// NewMonitor creates a new multicluster client monitor with Prometheus metrics.
+func NewMonitor() *Monitor {
+	return &Monitor{
+		crossClusterNameConflicts: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "cortex_multicluster_cross_cluster_name_conflicts_total",
```
PaulPickhardt: I think, we should make this configurable as every service that uses this package would have this name for the metric. This would collide then. What do you think?

PaulPickhardt: I could imagine adding an optional field to `NewMonitor()`

PhilippMatthes:
Waiting for pauls review

pkg/multicluster/monitor.go
PaulPickhardt:
```
func NewMonitor(namespace string) *Monitor {
    prometheus.NewCounterVec(prometheus.CounterOpts{
        Namespace: namespace,
        Name:      "multicluster_cross_cluster_name_conflicts_total",
```

PaulPickhardt (changes requested):
Other then that looks really nice! Thanks for your effort!

PaulPickhardt:
One could also argue about having an option for hard constraining not to create CROs when there is an error other then the duplicate one.

But I guess as you already wrote "best effort" it is a tradeoff between one cluster/namespace could block everything and having a duplicate because of one cluster having an issue.

So I guess this is fine.

PaulPickhardt (approved):

SoWieMarkus (approved):

umswmayj (approved):

PR #1125, opened 2026-08-07
feat: CR controller checks for host overload scenario and resolves via reservation re-placements

mblos:
This PR adds automated detection and remediation when a host accumulates more committed reservation blocks than its actual capacity allows: Cortex now evicts reservation slots from oversubscribed hosts so that committed capacity guarantees can be honored.

A grace period is enforced before eviction to allow other controllers (e.g. failover) to self-heal first. Operators can enable detection-only mode to observe violations via metrics before turning on active remediation.

- Automatically restore capacity headroom on oversubscribed hosts
- Avoid unnecessary evictions during transient spikes via configurable grace period
- `enableOversubscriptionUnplaceReservations: false` reports violations without evicting


internal/scheduling/reservations/commitments/reservation_controller.go
```diff
@@ -863,3 +909,249 @@ func (r *CommitmentReservationController) SetupWithManager(mgr ctrl.Manager, mcl
...
+	})
+	memViolation := violations[hv1.ResourceMemory]
+	for _, res := range allocatedReservations {
+		unused := reservations.UnusedReservationCapacity(res, false)
+		unusedMem := unused[hv1.ResourceMemory]
+		if unusedMem.Cmp(memViolation) >= 0 {
+			allocatedReservations = []*v1alpha1.Reservation{res}
+			break
+		}
+	}
```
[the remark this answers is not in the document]
mblos: false positive. if pushing away used reservations, the intent is to keep the VMs on the host, but pushing away the reservations -> reset allocations of that reservations

helm/bundles/cortex-nova/values.yaml
```diff
@@ -185,6 +185,16 @@ cortex-scheduling-controllers:
...
+      # How long to wait after detecting host over-subscription before evicting reservation slots. 
+      # Gives other controllers (e.g. failover) time to self-heal.
+      oversubscriptionGracePeriod: "3m"
+      # Minimum time between consecutive over-subscription checks for the same host.
+      # Should be shorter than oversubscriptionGracePeriod.
+      oversubscriptionMinCheckInterval: "1m"
+      # Enable host over-subscription detection
+      enableOversubscriptionCheck: true
+      # When false, violations are detected and exposed via metrics but no slots are evicted.
+      enableOversubscriptionUnplaceReservations: true
```
juliusclausnitzer: Not a big fan of this naming since `unplace` is not a real verb (at least in this century :)). How would you like `enableOversubscriptionEviction`,  `enableReservationEviction` or `enableOversubscriptionRemediation` ?

internal/scheduling/reservations/commitments/api/report_capacity_test.go
```diff
@@ -250,10 +250,12 @@ func TestCapacityCalculator(t *testing.T) {
 				ramRes := report.Resources["hw_version_test-group_ram"]
 				if ramRes == nil {
 					t.Fatal("missing hw_version_test-group_ram")
+					return
 				}
 				az := ramRes.PerAZ[tc.checkAZ]
 				if az == nil {
 					t.Fatalf("missing entry for AZ %s", tc.checkAZ)
+					return
```
juliusclausnitzer: why do we need all of these empty returns in otherwise untouched tests?

internal/scheduling/reservations/commitments/host_oversubscription_controller.go
```diff
@@ -0,0 +1,400 @@
...
+		return gracePeriod, nil
+	}
+	target := selectEvictionTarget(ctx, allReservations, violations)
+	if target == nil {
+		logger.Info("no evictable slot found — manual intervention required, will retry after grace period")
+		r.resetFirstSeen(host)
+		return gracePeriod, nil
+	}
+
+	if _, err := unplaceReservation(ctx, r.Client, target); err != nil {
```
juliusclausnitzer: same naming here

internal/scheduling/reservations/commitments/host_oversubscription_controller.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+	return violations
+}
+
+// selectEvictionTarget picks the best slot to evict from allReservations given the active violations.
+func selectEvictionTarget(
+	ctx context.Context,
+	allReservations []v1alpha1.Reservation,
+	violations map[hv1.ResourceName]resource.Quantity,
+) *v1alpha1.Reservation {
```
juliusclausnitzer: The `violations` map is passed into `selectEvictionTarget` but only used for logging — the selection logic ignores it entirely. This means we always evict the smallest unallocated slot regardless of whether it actually covers the violation, potentially requiring multiple grace period cycles to resolve what a single well-chosen eviction could fix.

Consider picking the smallest slot whose resources are `>= violations` before falling back to the global smallest. Would reduce unnecessary wait time on hosts that are significantly oversubscribed.

At the least we should adjust the comment, since we do not try to find the best slot and do not use the active violations (except for logging).

mblos: lets discuss live, i also thought about what is better.. multiple eviction or very small slots, or a few or single eviction of a larger slot. the latter is harder to find a new host for - thats why its currently always trying to prefer smallest slots to evict. so it needs multiple runs, but converge eventually

internal/scheduling/reservations/commitments/api/report_capacity_test.go
mblos: was some auto linting thing, removed it

internal/scheduling/reservations/commitments/host_oversubscription_controller.go
mblos: kiss for now

SoWieMarkus (approved):

juliusclausnitzer (approved):

PR #1135, opened 2026-08-12
feat: add monitor for caching client

SoWieMarkus:
still need to test that locally. tilt setup is currently borked


SoWieMarkus:
Closing this since the refactor completely changed the way the monitor needs to be setup :D

PR #1160, opened 2026-08-25
feat: add cache monitoring

SoWieMarkus:
Adds a high-watermark size metric for the pending-cache overlay. Because overlay entries are evicted within milliseconds, a plain gauge would almost always read zero and give no signal. So `cache.Monitor` tracks the peak entry count per (gvk, host) between scrapes and resets it on each Collect, alongside the current size at scrape.

- `cortex_cache_overlay_entries_max{gvk,host}` - high-watermark since last scrape (reset on scrape)
- `cortex_cache_overlay_entries{gvk,host} ` - current entry count

With that we can track that if the eviction is working as intended or if there is a memory leak

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -393,8 +393,11 @@ func main() {
 	}
 	multiclusterClientConfig := conf.GetConfigOrDie[multicluster.ClientConfig]()
 
+	var cacheMonitor *cache.Monitor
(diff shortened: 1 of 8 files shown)


PhilippMatthes (commented):
Can we already define an alert?

PhilippMatthes (dismissed):

PhilippMatthes (approved):

